PR #13, opened 2017-01-01
Reverse string Solution

MJUIUC:
wrote a quick sol in python.


jarydremillard:
Reviewed. Correct path.
Merging PR.

PR #20, opened 2017-01-02
php/selectivealso solution for the string reversal

SelectiveAlso:
There's a long and a short solution. Check the readme.


jarydremillard:
Reviewed. No conflicts and path is correct. Merging PR.

PR #36, opened 2017-01-02
Adding solution

JRTN:
solution


jarydremillard:
Hi JRTN.

Good job. Just a note that instead of creating uninitialized variables on a single line, you can cut your program almost in half. so instead of 
`int i;
for (i =0;...`

you can do
`for (int i = 0;...`

This goes for
`char temp;
temp = ...`

can do
`char temp = ...`

That literally cuts the code in half.

Other than that good job.

Code reviewed. No conflicts.
Merging PR.

PR #38, opened 2017-01-02
First challenge

GShooty:


jarydremillard:
It might be good practice to specify what version of python you are using, otherwise this wouldnt run on a machine with 3.x installed!

Reviewed. No conflicts.
Merging PR.

PR #39, opened 2017-01-02
whatever you want

MicalV:


jarydremillard:
Interesting use of end=""

Might be good to add comments to your function so readers will have an instant idea of how you reverse the string.

I would also avoid using single character variables. In this case, not too big of a deal since its a small challenge. But in production, it can get messy, or more difficult challenges can get confusing!

Good code!

- [x] Reviewed
- [x] No conflicts
- [x] **Merged PR**

PR #40, opened 2017-01-02
solution for challenge_1 in c++

ManuelMeraz:


jarydremillard:
- [ ] Code review
- [ ] Correct path
- [ ] Check for conflicts
- [ ] **Merge PR**

jarydremillard:
Code review:

Using `using namespace std;` would save you a lot of typing keystrokes. In the short run, it doesn't mean anything, but in the long run, you could be saving thousands of keystrokes.

At line 18, you do a conditional on quit. The conditional would be triggered if its true by default, therefore, setting it to true when its triggered doesn't really do anything and is a waste of memory.


Otherwise, good use of index!

PR #66, opened 2017-01-02
Solves Challenge_2

hkl0902:


josephpd3:
Nice!

You may be interested in [Counters](https://docs.python.org/3/library/collections.html#collections.Counter). It's a Dict made specially for this sort of thing, and it is one of many really cool tools in the `collections` module that is included in standard Python.

PR #70, opened 2017-01-02
Lonely Characters Club (Challenge 2 Solution in Rust)

josephpd3:
Docs, Source, and Tests included!

### DIFF
--- a/challenge_2/rust/josephpd3/src/lib.rs
+++ b/challenge_2/rust/josephpd3/src/lib.rs
@@ -1,0 +1,33 @@
+use std::collections::HashMap;
+
+pub fn get_lonely_items(v: &[char]) -> Vec<char> {
+    let mut lonely = vec![];
+    let mut counts = HashMap::new();
+
+    for ch in v {
+        let mut present = true; // Flags are fun!
+
+        match counts.get_mut(ch) {
+            // Get mutable reference to entry
+            Some(c) => { // If entry exists...
+                *c = *c + 1; 
+            },
+            None => { // If entry doesn't exist...
+                present = false;
+            }
+        }
+        
+        if !present { // Add base entry if didn't exist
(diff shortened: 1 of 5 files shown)


challenge_2/rust/josephpd3/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+        }
+    }
+
+    for (key, val) in counts.iter() {
+        if *val as u32 == 1u32 { // Gotta deref value
+            lonely.push(**key);
+        }
+    }
+
+    lonely.sort(); // Return in lexicographical order
```
myrrlyn: I don't remember the sort time complexity offhand, so I'm curious as to whether you're still in linear time with this included

challenge_2/rust/josephpd3/src/lib.rs
```diff
@@ -0,0 +1,33 @@
+use std::collections::HashMap;
+
+pub fn get_lonely_items(v: &[char]) -> Vec<char> {
+    let mut lonely = vec![];
+    let mut counts = HashMap::new();
+
+    for ch in v {
+        let mut present = true; // Flags are fun!
```
myrrlyn: Even more fun than flags are Rust's error chaining. I'm reviewing this from the airport so I can't fact-check myself, but would a `Result` returning accessor like `.get(ch).or_insert(ch, 0)` work here?

challenge_2/rust/josephpd3/src/lib.rs
josephpd3: @myrrlyn Good catch! I actually glazed over this myself. Thanks for reviewing my PR! I'm still working on figuring out exactly how to make my Rust more idiomatic, and it's great to have someone else here to look at my code.

PR #73, opened 2017-01-02
Create challenge_1

Jzanet:
SImple code to invert a given list using the list(), input() and .reverse() functions

### DIFF
--- a/challenge_1/python/zanetti/challenge_0.py
+++ b/challenge_1/python/zanetti/challenge_0.py
@@ -1,0 +1,5 @@
+a=list(input("Insert the  you want to invert :")) #input to user choose the characteres to invert - in a list form
+
+print('the choosen world: ', a) 
+a.reverse() #invert the "a" list
+print('reversed: ', a)


ManuelMeraz:
- [ ] Code review
- [ ] Correct path
- [ ] Check for conflicts
- [ ] **Merge PR**

ManuelMeraz:
It looks like your code does not work correctly. You're almost there though! Just convert the list back to a string.

Here's the output I get with the input 'test'
```
Insert the  you want to invert :test
the choosen world:  ['t', 'e', 's', 't']
reversed:  ['t', 's', 'e', 't']
```

jarydremillard:
Code Review:

Technically, this works. But using user input for this case is iffy, especially since there was no check for integrity. Besides, not only are you limiting yourself to just a string, but if you actually set it to a variable, you could reverse strings and lists. Other than that, way to keep it minimal! I like concise code!

Jzanet:
Ok, Thanks, I'll correct them as soon as possible.

PR #76, opened 2017-01-02
made changes to solutions 2 and 3

MicalV:


challenge_2/python/mical/main.py
```diff
@@ -0,0 +1,12 @@
+#02/01/2017
+
+def noRepeats(list):
+    for x in list:
+        amount = list.count(x)
```
moonpatroller: Isn't count O(n)?  That would make entire solution O(n^2).

jarydremillard: Isnt `list` a python reserved keyword? Regardless... Naming a variable after a data structure isn't the best habit. I would advise to change it to something more accurate! Otherwise, good job!

MicalV:
Didn't notice there was a comment thing here. I changed the variable names, hopefully this is good

PR #92, opened 2017-01-02
Create main.py

wOstensen:
Finds all single values from a given list. :)

### DIFF
--- a/challenge_2/python/wost/main.py
+++ b/challenge_2/python/wost/main.py
@@ -1,0 +1,32 @@
+"""
+Python 3.6:
+  :: Counts all the instances of all the elements in a list.
+  :: Returns all the instances with a count of 1.
+"""
+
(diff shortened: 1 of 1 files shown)


ASethi77:
- [ ] **Code Correctness**
No issues with code correctness. Good job on your test suite!
- [ ] **Code Style**
No issues with code style.
- [ ] **Code Documentation**
Approach is not documented. Recommend writing a quick README that describes your approach.
- [ ] **Other**
Just a side note: I understand that this README is somewhat silly for this challenge, but we would like people to get used to this habit early, or else documentation will fall apart in challenges and projects where this is much more important. We'd like the community to establish good coding practices to make the most out of this project. I'm just recommending this for now but in the future we may be more strict on this requirement.

jarydremillard:
Bypass

PR #93, opened 2017-01-02
Java Challenge 2 (Pending)

Sphinx111:
Submitting solution to challenge_2 in java.

Didn't manage to run in Linear complexity but still happy with the solution. Code also generates randomised array for testing.

### DIFF
--- a/challenge_2/java/eddy_north/findLoneNumber.java
+++ b/challenge_2/java/eddy_north/findLoneNumber.java
@@ -1,0 +1,88 @@
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Random;
+
+public class findLoneNumber {
+
+	public static void main(String[] args) {
+		String[] sourceArray = generateRandomisedArray(26,true);
+		String answer = findLoneNumber(sourceArray);
+		System.out.println(answer);
+	}
+	
+	public static String[] generateRandomisedArray(int elements,boolean useCharacters) {
+		boolean charactersIncluded = useCharacters;
+		int n = elements;
+		if (n % 2 == 0) {
+			n = n+1;
+		}
+		String loneString = "";
+		String[] objectArray = new String[n];
+		boolean singleValueAdded = false;
+		for (int i = 0; i < n; i++) {
+			String includeThisValue = "" + (int)Math.floor((Math.random()*n));
+			while (includeThisValue.equals(loneString)){
+				includeThisValue = "" + (int)Math.floor((Math.random()*n));
+			}
+			if (charactersIncluded) {
+				//if problem using characters, introduce 50% chance to overwrite int with char.
+				if (Math.random() < 0.5) {
+					Random r = new Random();
+					char c = (char)(r.nextInt(26) + 'a');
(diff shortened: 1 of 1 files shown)


challenge_2/java/eddy_north/findLoneNumber.java
```diff
@@ -0,0 +1,88 @@
...
+		for (int i = 0; i < sourceArray.length; i++) {
+			if (!countElements.containsKey(sourceArray[i])) {
+				countElements.put(sourceArray[i], 1);
+			} else {
+				int newValue = countElements.get(sourceArray[i]) + 1;
+				countElements.put(sourceArray[i], newValue);
+			}
+		}
+		//Following is O(n^2) solution, O(n) solution not finished yet :(
+		for (int i = 0; i < countElements.size(); i++){
```
moonpatroller: This outer loop looks unnecessary.

challenge_2/java/eddy_north/findLoneNumber.java
```diff
@@ -0,0 +1,88 @@
...
+
+public class findLoneNumber {
+
+	public static void main(String[] args) {
+		String[] sourceArray = generateRandomisedArray(26,true);
+		String answer = findLoneNumber(sourceArray);
+		System.out.println(answer);
+	}
+	
+	public static String[] generateRandomisedArray(int elements,boolean useCharacters) {
```
moonpatroller: You should be able to generate the array in 1/4 the code or less.

ASethi77:
- [ ] **Code Correctness**
Random test cases are good, but not as a set of initial test cases. You should explicitly define edge cases/situations you want to handle and show via deterministic tests that you handle them. Randomised tests open up a can of worms in the future for ensuring that you've actually tested all possible cases, and if those cases/requirements ever change in the future. I also noticed that your code seems to change the number of elements in the array if the user of the function requested an ever number of elements.
- [ ] **Code Style**
Variable names could be improved (e.g. includeThisValue) and spacing is not always consistent, especially between operators. Several nested loops + conditional statements make it somewhat difficult to understand control flow between cases
- [ ] **Code Documentation**
Code is not documented. When I say documented, it would be especially helpful to have some more inline comments that sort of trace through your algos, such as what case a certain code branch is trying to handle. Would recommend documenting approach via README as well, especially for generating randomised array (very cool btw).
- [x] **Other**
Just want to let you know that we're being somewhat light on documenting for early challenges, but that we'll be more strict on this in the future. The intent of this is to help you establish habits that are critical to any successful medium or large project, or when you have to read your own code much later :)
- [ ] **Merge PR**

ManuelMeraz:
I agree with @ASethi77 . I think you should document your code and explain your thought process behind your code. Following that, I think as long as you meet the criteria for solving the problem, even if it might not be optimal, you should be fine. Thanks!

I've changed the status of your Pull Request to pending. Once you've added some documentation, we can pull it in, simply remove the pending status from the title by editing it.

ManuelMeraz:
I've changed the status of your pull request to pending. Once you've made the changes you wanted to your code, simply remove the pending status from the title.

PR #95, opened 2017-01-02
Challenges/rawphs

Rawphs:
Challenges 0-3


jarydremillard:
- [ ] Code review
- [ ] Correct path
- [ ] Check for conflicts
- [ ] **Merge PR**

Rawphs:
@jarydremillard I am using node just because I wanted to use the command line instead of the browser, but if I should write it using vanilla js I can rewrite the challenges and add an .html, no problem. Semicolons are optional for Node.js because it also has automatic semicolon insertion.

myrrlyn:
Node should be fine; requiring an HTML display for challenges we're running primarily in the console on other projects seems to me like it'd be an unfair handicap.

challenge_2/javascript/rawphs/index.js
```diff
@@ -0,0 +1,13 @@
...
+function findUnique (array) {
+	return array.filter(
+		item => array.indexOf(item) === array.lastIndexOf(item)
+	)[0]
+}
+
+console.log(findUnique(testArray1))
+console.log(findUnique(testArray2))
+
+if (process.argv[2]) console.log(findUnique(JSON.parse(process.argv[2])))
```
myrrlyn: It's considered good practice to put even (if not especially) one-liner conditional clauses inside curly-braces, just to prevent borrowing trouble in the future.

And that's the only nitpick I've been able to find with you so far. Good work.

Rawphs: I agree with you on that, but this is now how they do on my current job so I thought I was a bit behind on what is considered good practice or not. I also had to let go on this whole semicolon thing hahahahhahaha. Thanks for your review!

myrrlyn: semicolons schmemicolons; JavaScript does weird things. I really like being pedantic and putting in every symbol even remotely necessary, but I have a C background full of pain from eliding such things.

PR #97, opened 2017-01-02
C Challenge 1, 2 and 3 (Pending)

xoumi:


jarydremillard:
- [ ] Code review
- [ ] Correct path
- [ ] Check for conflicts
- [ ] **Merge PR**

challenge_2/c/xoumi/SingleDigit.c
```diff
@@ -0,0 +1,24 @@
...
+{
+    char array[255],i;
+    int flag,len,j;
+    printf("Enter array of numbers: ");
+    gets(array);
+    for(len=0;array[len]!='\0';len++);
+    len-=1;
+
+    //Run every number from 0 to 9 and if matching, flag+=1
+    //This has a runtime of O(11n) I think. That's linear right ?
```
moonpatroller: The example has ints like 10, so you should change this to handle that.  Your nested loops will be O(n^2).

challenge_3/c/xoumi/majority.c
```diff
@@ -0,0 +1,25 @@
...
+    int flag,len,j;
+    printf("Enter array of numbers: ");
+    gets(array);
+    for(len=0;array[len]!='\0';len++);
+    printf("%d \n",len);
+
+    //Run every number from 0 to 9 and if matching, flag+=1
+    //only changed one line from Challenge_2 at line20.
+    //This has a runtime of O(11n) I think. That's linear right ?
+    //would have a run time of O(37n) if I included characters.
```
moonpatroller: This also only handles single digit numbers.

challenge_2/c/xoumi/SingleDigit.c
xoumi: *facepalms
Well you're right, neither of these work for anything more than two digit numbers.
This entire algorithm fails if I cross single digit numbers...
back to drawing board I guess. Thanks.

challenge_1/c/xoumi/ReverseString.c
```diff
@@ -0,0 +1,27 @@
...
+
+    for(j=0;j<i/2;j++){
+	temp=string[j];
+	string[j]=string[i-j];
+	string[i-j]=temp;
+    }
+
+    printf("Reversed string is %s \n", string);
+}
+
```
halogenandtoast: What do you think about removing all the extra newlines at the bottom here?

challenge_1/c/xoumi/ReverseString.c
```diff
@@ -0,0 +1,27 @@
...
+void main()
+{
+    int i,j;
+    char *string,temp;
+    printf("Enter string: ");
+    //because scanf is scary
+    gets(string);
+
+    //Just a loop to count length instead of using strlen because this isn't scary
+    for(i=0;string[i]!='\0';i++);
```
halogenandtoast: Is there a reason not to use `strlen` I don't know of any reason it would be considered scary.

challenge_1/c/xoumi/ReverseString.c
```diff
@@ -0,0 +1,27 @@
+#include<stdio.h>
+void main()
```
halogenandtoast: I think it's more canonical to use `int` as the return type for which you return the exit status code. What are your thoughts about switching it to that.

xoumi: if this was part of a bigger more complex program then yes I would use int I guess. I didn't really have any special reason to use void, just did~
I can most certainly use int here :)

challenge_1/c/xoumi/ReverseString.c
xoumi: It's not :p
Just a fun way to count the length I guess and apparently strlen is basically just that according to someone in the slack channel, I can't recall who.

challenge_1/c/xoumi/ReverseString.c
xoumi: Well moonpatroller pointed out, this algorithm won't work for anything beyond single digit numbers so I'm gonna write from scratch anyway.

challenge_1/c/xoumi/ReverseString.c
ASethi77: Indeed strlen basically scans until it hits the 'null terminator' so this code is equivalent to using strlen. Just as a side note, `gets` is also scary and can be used in stack overflow attacks: check out [this stackoverflow thread](http://stackoverflow.com/questions/3302255/c-scanf-vs-gets-vs-fgets) for a  slightly more detailed description.

halogenandtoast: While you can write your own strlen for learning purposes you really shouldn't. Strlen can be optimized by the compiler to be much more efficient. Just take a look at the glibc implementation http://www.stdlib.net/~colmmacc/strlen.c.html

ASethi77:
- [ ] **Code Correctness**
Code for challenge 1 seems correct, but not for challenges 2 and 3. Obviously this is in work based on previous comments, I'm just documenting this for now :) I also encourage you to write a test-runner in your main that shows hard-coded test cases and maybe some asserts to show what expected output is, or to document exactly what functionality you tested in a README.
- [ ] **Code Style**
There are issues with spacing between operators in your code; I encourage you to take a look at [this example snippet](https://gist.github.com/davidzchen/9187984) which shows sample C code that adheres to the Linux Kernel style guidelines. These are commonly-accepted style patterns regarding spacing that I think your code can benefit from. Other comments regarding refactoring certain blocks are already mentioned above.
- [ ] **Code Documentation**
We're being really lax on this right now, but we encourage you to write a README that discusses your high-level approach for each challenge. We will likely enforce this requirement for future challenges.
- [ ] **Other**
- [ ] **Merge PR**

https://gist.github.com/davidzchen/9187984

PR #102, opened 2017-01-03
[Python] Challenges #0, #1, #2, #3, #4

ningyuansg:
Reviewers:

1. @moonpatroller in Challenge #2, implemented in 0600f3a
2. @hkl0902 in Challenge #3, implemented in 61d340a

Questions:

1. In Challenge #4, 94ee769, I created my own class of `BTree` to represent the binary tree. Is there some other better way, perhaps in-built, in python to represent a binary tree? For context, I have no education in computer science.


challenge_2/python/ning/challenge_2.py
```diff
@@ -0,0 +1,20 @@
+def find_unique(sequence):
+    item_counter = dict()
```
moonpatroller: You should just use [] and {} instead of list() and dict().

hkl0902:
Looks good! A way to speed Challenge_3 up (slightly) is instead of having a separate for loop to check whether `item_count` > `len(sequence)/2` you could do this in the first for loop. For every `item in sequence`, just check if `item_counter[item] > len(sequence)/2`. If it is, return that item. 

Also, python has a [Counter class](https://docs.python.org/dev/library/collections.html#collections.Counter) that just counts objects. Might be useful in the future.

challenge_2/python/ning/challenge_2.py
ningyuansg: Hi, thanks for your input. I previously used `list() or dict()` because I thought they were more readable. But considering that `[] or {}` are both more conventional and efficient, I will follow your advice.

I have also changed the code to use defaultdict in 0600f3a. I did not previously know about this module, but now I do.

ningyuansg:
@hkl0902 Hi, thanks for the suggestions! I have changed my solution in 61d340a to use the Counter class you mentioned. It is now effectively a one-liner.

jarydremillard:
- [ ] Code review
- [ ] Correct path
- [ ] Code tested
- [ ] Check for conflicts
- [ ] **Merge PR**

PR #112, opened 2017-01-03
[Python] Challenge 1, 2 and 3 (Pending)

alexbotello:
Completed.

### DIFF
--- a/challenge_2/python/alexbotello/main.py
+++ b/challenge_2/python/alexbotello/main.py
@@ -3,4 +3,5 @@ if __name__ == '__main__':
     array = [int(input()) for _ in range(N)] # Enter ints
 
     print("\nThe single number is ", end='')
-    print(*[i for i in array if array.count(i) == 1], sep=' or ')
+    #print(*[i for i in sorted(array) if array.count(i) == 1], sep=' or ')
+    print(min(array, key=array.count))
--- a/challenge_2/python/alexbotello/main.py
+++ b/challenge_2/python/alexbotello/main.py
@@ -1,0 +1,6 @@
+if __name__ == '__main__':
+    N = int(input()) # How many items in the list?
+    array = [int(input()) for _ in range(N)] # Enter ints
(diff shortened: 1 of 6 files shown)


challenge_2/python/alexbotello/main.py
```diff
@@ -0,0 +1,6 @@
+if __name__ == '__main__':
+    N = int(input()) # How many items in the list?
+    array = [int(input()) for _ in range(N)] # Enter ints
+
+    print("\nThe single number is ", end='')
+    print(*[i for i in array if array.count(i) == 1], sep=' or ')
```
zmiller91: This is really slow, it's O(N^2) because the array isn't sorted so the `count()` method has to iterate through every element in your array. It's essentially a doubly nested for loop.

alexbotello: Thanks for letting me know. I've been trying to practice list comprehensions and use them more in my code.
Would using the sorted function fix this?
`print(*[i for i in sorted(array) if array.count(i) == 1])`

alexbotello: Thanks for letting me know. I'm just learning list comprehensions trying to use them more often in my code.
Would using the sorted function in the comprehension remedy this?

zmiller91: Probably not; the method will still probably look through every element. What if you counted all the elements up front? Could you iterate over the counts and fine the element that showed up once?

ManuelMeraz:
- [ ] **Code Correctness**
- [ ] **Code Style**
- [ ] **Code Documentation**
- [ ] **Path Structure**
- [ ] **Merge PR**

alexbotello:
I do indeed get 6 as the most common number for challenge 3. I'm using Python 3.6 and taking advantage of the new order-preserving aspect of dict

ManuelMeraz:
I'll install python 3.6 and run it!

PR #113, opened 2017-01-03
[Python] Challenge 0,1,2, and 3 (Pending)

paper-squares:


Challenge_2/Python/Papersquares/Challenge2uniquevalue.py
```diff
@@ -0,0 +1,8 @@
+
+list = [2,3,4,2,3,5,4,6,4,6,9,'a',10,9,8,7,8,10,7]
+
+newlist = []
```
moonpatroller: Why not a list comprehension?

paper-squares: I don't know what 0(n^2) is. 

in terms of list comprehension I had to do some research but I figured out

newlist = [ ITEM for ITEM in LIST if LIST.COUNT(ITEM) ==1]

paper-squares: What benefit does this have in regards to things?  Just because it is shorter code it makes this run faster or does it have other functionality?

ManuelMeraz:
- [ ] **Code Correctness**
- [ ] **Code Style**
- [ ] **Code Documentation**
- [ ] **Path Structure**
- [ ] **Merge PR**

Challenge_2/Python/Papersquares/Challenge2uniquevalue.py
ManuelMeraz: Hey there. What @moonpatroller is saying is that you are iterating over the list for every index in that list, thus making it O(n^2). 

Shorter code, does not necessarily mean faster in terms of processing.

think about the function list.count() and how would you code it? To count a value you have to iterate over the whole list to determine the number of times a value is repeated.

So you're iterating through your list, and for every value you are iterating through the whole list again to count each value, thus you are performing len(list) * len(list) operating, where by len(list) is the `n` he is referring to in o(n^2) if that makes sense.

PR #116, opened 2017-01-03
Solution to challenges 0-4 (python)

Atakyn:


challenge_3/python/Atakyn/c3.py
```diff
@@ -0,0 +1,18 @@
+# find majority element in array
+
+
+def find_majority_element(array):
+    majority_value = len(array) // 2
+    values = {}
+    for e in array:
+        try:
+            values[e] += 1
```
moonpatroller: Why not use defaultdict?

challenge_4/python/Atakyn/c4.py
```diff
@@ -0,0 +1,33 @@
...
+
+def reverse(root):  # solution, accounts for unpaired nodes (e.g. if 10 were added to the challenge example tree)
+    if root is not None:
+        if root.left is not None:
+            temp = root.left
+            root.left = root.right
+            root.right = temp
+            reverse(root.left)
+            reverse(root.right)
+        elif root.right is not None:
```
moonpatroller: I think this should be an if, not elif.

challenge_3/python/Atakyn/c3.py
Atakyn: Didn't know about that - switching to using it now

challenge_4/python/Atakyn/c4.py
Atakyn: Fair enough - fixed

MJUIUC:
Looks good. Awesome job reviewing and writing!

PR #117, opened 2017-01-03
First submission

erocs:


jarydremillard:
- [ ] Code review
- [ ] Correct path
- [ ] Code tested
- [ ] Check for conflicts
- [ ] **Merge PR**

challenge_2/go/erocs/find_singleton.go
```diff
@@ -0,0 +1,29 @@
+package main
+
+import (
+	"fmt"
+)
+
+func FindSingleton(int_array []int) (found int) {
+	singles := make(map[int]int)
+	repeats := make(map[int]int)
```
natdm: Any reason you can't just have one map of `map[int]int` and just loop over it at the end to find the (or a) single and return that? One less map allocation?

challenge_2/go/erocs/find_singleton.go
```diff
@@ -0,0 +1,29 @@
...
+			repeats[n]++
+		} else if _, ok := singles[n]; ok {
+			delete(singles, n)
+			repeats[n] = 2
+		} else {
+			singles[n] = 1
+		}
+	}
+	for k, _ := range singles {
+		found = k
```
natdm: Might as well exit the for loop when you find one since you'll just be reassigning `found` each time you find one and it doesn't look like it matters what one you return.

challenge_3/go/erocs/majority.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import (
+	"fmt"
+)
+
+func FindMajority(int_array []int) int {
```
natdm: Go idioms is to use smaller naming conventions. This is also a slice, not an array. Maybe `ints`

challenge_4/go/erocs/invert_tree.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+type TreeNode struct {
+	Left  *TreeNode
+	Right *TreeNode
+	Value int
+}
+
+func (t *TreeNode) String() string {
+	l := ""
```
natdm: I think you'd more typically see 
```go
var l string
var r string
```

You could do the same for the numbers as above but it's mainly in strings where it's a bit odd to use shorthand to create an empty value string when it's already that way with `var`

challenge_2/go/erocs/find_singleton.go
erocs: This is one less loop iteration at the cost of one allocation. Different trade-offs. Realistically, these maps are being used as Sets and _repeats_ doesn't need to bother counting.

challenge_2/go/erocs/find_singleton.go
erocs: If I had the return within the loop, I would still need the end the function return, since the loop might contain nothing. It doesn't gain anything since this loop should only iterate one value, per spec. If I updated to have both return statements, I would also change the return value to (int, error) and rework the loop so an error could be returned on no single or multiple singles. Not adding the error return was just my laziness.  ;)

challenge_3/go/erocs/majority.go
erocs: _ints_ would be good as it is still descriptive enough.

challenge_4/go/erocs/invert_tree.go
erocs: Either works fine. I find explicit initialization to be more descriptive but each to his own.

erocs: (Also yes I do appreciate the irony of 1-letter variable names when talking about descriptiveness. Once again, laziness in my part to crank out a simple problem, lol. I'm glad code standards will be implemented in the repo.)

PR #120, opened 2017-01-03
Challenges 0, 1, 2 with Python

ajschrier:


ajschrier:
Almost forgot to mention the change to .gitignore, if you didn't notice

MJUIUC:
Interesting solution to challenge2. I did something similar. I see that you sort of made a histogram with two lists which is good. I'm not sure how long it takes python to check your entire lists. This is something that you can look up! I'm assuming it does not meet O(n) because it has to check through the entire list, though I could be wrong since list in python sort of work like hash table's anyway.

PR #124, opened 2017-01-03
[Java] Challenge 1 & 2 (Pending)

leighflix:


MJUIUC:
Could've sworn I just wrote something here... Anyway, nice job! Thanks for re-submitting with the correct file names! I also used a histogram in my solution!

PR #131, opened 2017-01-03
ndsmith3's first submissions

ndsmith3:


challenge_0/scala/minolwa/HelloWorld.scala
```diff
@@ -0,0 +1,3 @@
+object HelloWorld extends App {
```
zmiller91: what's up with the different file path? please remove.

ndsmith3: Whoops. Forgot to rm that directory. Consider it done.

PR #137, opened 2017-01-03
[Python] Challenge 4 (Pending)

fab0l1n:
added `print_tree()` to check solution

### DIFF
--- a/challenge_4/python/fab0l1n/Challenge4_solved
+++ b/challenge_4/python/fab0l1n/Challenge4_solved
@@ -1,0 +1,34 @@
+class Node:
+    def __init__(self, id, left=None, right=None):
+        self.id = id
+        self.left = left
+        self.right = right
+
+    def invert(self):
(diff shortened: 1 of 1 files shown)


wOstensen:
You could try ```self.left, self.right = self.right, self.left``` in your invert method, it does the exact same thing, but in one line. :)

MJUIUC:
I agree with what @popcornanachronism said. You're recursive calls need to be changed.

ManuelMeraz:
- [ ] Code review
It looks like you're trying to call the reverse() method on your node. Change it to invert(). 

Try and looking at a recursive wave to print out your tree that uses the properties of a binary tree. Loop up pre order traversals.
- [ ] Code documented
See if you can add some writing explaining your thought process it should help you think about your solution a bit more if you can explain it.
- [ ] Correct path
Good
- [ ] Check for conflicts
- [ ] **Merge Pull Request**

PR #139, opened 2017-01-03
Challenge 3 and 5 in Python!

wOstensen:
I have commited challenge 2 and 3, with both README's. :)


wOstensen:
Thank you so much! :D

PR #142, opened 2017-01-03
Challenge 1 in Python

slansford:


wOstensen:
You can do the input function inside the print function, that way you'll have one single line.

MJUIUC:
Yea this looks pretty solid. I didn't know you could write inputs that way so I learned something!

challenge_1/python/slansford/src/challenge_1.py
```diff
@@ -0,0 +1,9 @@
+#Program asks user for word, reverses word, and returns the reversed word
+
+#Asks user for input and reverses letters of word
+
+rev_word = input('What word would you like reversed? ')[::-1]
```
jarydremillard: Good use! Efficient! Its exactly what I wanted to see for this challenge.

challenge_1/python/slansford/README.md
```diff
@@ -0,0 +1 @@
+Prints reversed string provided by user input. Uses Python 3.6
```
jarydremillard: Thank you for including the version!

PR #148, opened 2017-01-03
Challenge 1 and 2 in python

nebi:


MJUIUC:
Your string reverse is good, however your unique character runtime is O(n^2). Try to think about how you could get it to O(n) because it is indeed possible.

nebi:
Thank you for your feedback ,  I have update my code to use O(n) instead of O(n^2) .

challenge_2/python/nebi/src/single_number.py
```diff
@@ -0,0 +1,13 @@
...
+
+def single(array):
+ y = []
+ for x in array:
+  if array.count(x) == 1:
+   y.append(x)
+ return (y)
+
+
+print(single(array = [2,'a','l',3,'l',4,'k',2,3,4,'a',6,'c',4,'m',6,'m','k',9,10,9,8,7,8,10,7,2,3,4,2,3,5,4,6,4,6,9,10,9,8,7,8,10,7]))
```
jarydremillard: this returns `[]` an empty list. this should return `c`.

`>>> def single(array):
...   y = []
...   for x in array:
...     if array.count(x) == 1:
...       y.append(x)
...     return (y)
... 
>>> print(single(array = [2,'a','l',3,'l',4,'k',2,3,4,'a',6,'c',4,'m',6,'m','k',9,10,9,8,7,8,10,7,2,3,4,2,3,5,4,6,4,6,9,10,9,8,7,8,10,7]))
[]
>>> print(single(array))
[]
`

MJUIUC: I ran it and it works ok. 
screen shot 2017-01-04 at 9 03 44 pm

PR #152, opened 2017-01-03
Challenge_2 in Python

polarbear529:


MJUIUC:
I agree with @OkraArmband . Even though this works, The larger the inputs get, the slower this will run. And that runtime will increase in O(n^2) magnitude. Try to stay away from that.

polarbear529:
@OkraArmband Thanks for your input! How does it run through the array 324 times? The for loop isn't running through the array, its running through the range of the array. After 6 instances of the count function being called, (1 loop through the array per call), the program finds five, and thus terminates. What am I missing here?

ManuelMeraz:
Closed by request.

PR #153, opened 2017-01-03
[Python 3] Challenges 3, 4 and 5 (Fixed changes, request review)

popcornanachronism:


challenge_3/python/popcorn_anachronism/src/challenge_3.py
```diff
@@ -0,0 +1,10 @@
+a = [2,2,3,7,5,7,7,7,4,7,2,7,4,5,6,7,7,8,6,7,7,8,10,12,29,30,19,10,7,7,7,7,7,7,7,7,7]
+
+def majority_element(array):
+    n = len(array)
+    for item in array:
+        if array.count(item) > int(n/2):
```
zmiller91: `array.count()` iterates through the entire list; that line of code plus the outer for loop makes this solution unscalable.

popcornanachronism: yes, that wasn't ideal at all - thank you for pointing it out. I have updated the code using a dict to create a frequency array, I think that is more efficient now.

challenge_3/python/popcorn_anachronism/src/challenge_3.py
```diff
@@ -0,0 +1,16 @@
+a = [2,2,3,7,5,7,7,7,4,7,2,7,4,5,6,7,7,8,6,7,7,8,10,12,29,30,19,10,7,7,7,7,7,7,7,7,7]
+
+def majority_element(array):
+    n = len(array)
+    array_freq = {}     #dict of frequencies of items in array
```
hkl0902: I would use the [Counter](https://docs.python.org/dev/library/collections.html#collections.Counter) class here. Keep in mind that using `item in array` means that you're reading through that array every time. This is trivial for small lists, but for large ones, you're implementation may approach growing at O(n^2) instead of just O(n)

challenge_3/python/popcorn_anachronism/src/challenge_3.py
```diff
@@ -0,0 +1,16 @@
...
+    for item in array:
+        if item in array_freq:  #if item already present in dict, increment by 1
+            array_freq[item] = array_freq[item] + 1
+        else:                   #otherwise add item to dict with frequency of 1
+            array_freq[item] = 1
+    for item in array_freq:     #check for item with majority frequency
+        if array_freq[item] >= int(n/2):
```
hkl0902: Are you sure it's greater than or equal? The majority element is the element that appears **more** than ⌊ n/2 ⌋ times.

challenge_3/python/popcorn_anachronism/src/challenge_3.py
popcornanachronism: I don't think my old solution was O(n^2), I ran a test case with 3570000 list elements and it gave me
time_oldversion: 0.400 seconds
time_newversion: 0.153 seconds

Regardless, Counter was more efficient and I've added that with my new commit. Thanks!

challenge_3/python/popcorn_anachronism/src/challenge_3.py
popcornanachronism: Good catch, got that.

PR #154, opened 2017-01-03
Challenge 0

briguy-official:
boilerup_brian, challenge 0 in C


jarydremillard:
Good job include `return 0`! Lots of people forget about this! Very C like!

Bypass

PR #155, opened 2017-01-03
Created my Hello World program

BlakeBarnes00:
Using make to build my programs, I am on Ubuntu so it is best

### DIFF
--- a/challenge_0/cpp/blakebarnes00/Makefile
+++ b/challenge_0/cpp/blakebarnes00/Makefile
@@ -1,0 +1,5 @@
+CXX = g++
+FLG = -w -std=c++11
+
+all:
+	$(CXX) $(FLG) -o HelloWorld main.cpp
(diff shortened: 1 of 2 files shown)


jarydremillard:
- [ ] Code review
- [ ] Correct path
- [ ] Code tested
- [ ] Check for conflicts
- [ ] **Merge PR**

MJUIUC:
Good job! I'm not sure if you're aware, but "using namespace std" will get rid of the need for std before you cout. It's not important, but just a tip if you didn't know.

jarydremillard:
bypass

BlakeBarnes00:
@MJUIUC I am aware of that, just it is bad practice in theory. When you have bigger projects you want to be able to identify what you're using where, albeit that this is a simple program using the standard library and all, I would rather keep that in my mind as I learn C++ more.

Thank you for the heads up though!

PR #156, opened 2017-01-04
challenge 0

Slickness:


challenge_0/python/slickness/README.MD
```diff
@@ -0,0 +1 @@
+Hello world with python 3.5
```
jarydremillard: Thank you for including version of python. In the future, try to add how to run this as well! And include your thought process on completing a challenge. Although obviously it is not needed here for this challenge, but more or less this is for future reference.

Thank you.

jarydremillard:
Bypass

PR #158, opened 2017-01-04
challenge_1 in cpp

BlakeBarnes00:
Ignore the challenge_2 in cpp, I accidentally pressed tab twice and wasn't aware at the time (My ssh to my Raspberry Pi had lagged a little).

I successfully reversed a complete string using the algorithm header from standard.


MJUIUC:
Nice. I can't exactly ignore your ch2. You'll have to re submit with the updated one.

PR #159, opened 2017-01-04
Challenge 3, 4, and 5 in C++

ManuelMeraz:


MJUIUC:
Yup your logic makes sense to me and your code is clean and readable.

PR #165, opened 2017-01-04
[Python] Challenge 0-2 (Unreviewed)

bryantpq:


challenge_1/python/bryantpq/reverse.py
```diff
@@ -0,0 +1,5 @@
+def reverse_str(s):
+    print(s[::-1])
```
hkl0902: Might be a 'better' solution to return `s[::-1]` instead of simply printing it.

challenge_2/python/bryantpq/search_item.py
```diff
@@ -0,0 +1,24 @@
+def search_item(a):
+    d = {}
+    for i in a:
+        if i in d:
+            d[i] += 1
```
hkl0902: There's a [Counter](https://docs.python.org/2/library/collections.html#collections.Counter) class just for this.

challenge_3/python/bryantpq/find_majority.py
```diff
@@ -0,0 +1,17 @@
+def find_majority(a):
+    d = {}
+    l = len(a)
+    for i in a:
+        if i in d:
+            d[i] += 1
+            if d[i] >= l / 2:
```
hkl0902: If `d[i] = 10` and `l = 20` is `i` the majority element? Remember that "The majority element is the element that appears **more** than ⌊ n/2 ⌋ times."

challenge_4/python/bryantpq/tree.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+    def invert(self):
+        self.root = self._invert(self.root)
+
+
+    def _invert(self, node):
+        # find lowest point where nodes can be swapped
+        # swap nodes
+        if not (node.l.l == None and node.l.r == None): # node.left is terminal/leaf
```
hkl0902: This is a bit of a mouthful. 

In essense you're checking $not (not node.l.r and not node.l.r)$. This can be boiled down to $node.l.l or node.l.r$ which doesn't actually check if `node.left` is a leaf or not.  In fact, it checks if `node.left` isn't a leaf. Also, keep in mind that if `node.l` is None, you're going to get an error.

hkl0902:
Solutions 0-2 look good.

The error in Challenge 3 should be easy to fix. 

However, Challenge 4 is not a correct solution and will require a litle bit of more thought. Try to not add more if statements.

challenge_1/python/bryantpq/reverse.py
jarydremillard: Good point @hkl0902 thank you for your review!

jarydremillard:
I agree with @hkl0902 

I unfortunately have to set this in review till the fixes are committed and pushed.

@bryantpq If you have any questions, feel free to join our Slack channel and ask, or comment here.

challenge_3/python/bryantpq/find_majority.py
bryantpq: I see what's wrong. Thanks!

bryantpq:
@jarydremillard I have fixed challenges 3 and 4. Is it possible to update this pull request such that I don't have to make a new pull request every time I make a fix? I have already pushed the changes to my forked repo. Sorry I'm still new to git. Thanks in advance!

jarydremillard:
@bryantpq Ill approve.

PR #166, opened 2017-01-04
Pull Request

lex32pem:


challenge_3/python/lex32pem/challenge_3.py
```diff
@@ -0,0 +1,58 @@
...
+    print ("Here is your Array: " , userArray)
+
+majArray = find_Max(userArray)
+
+if majArray[1] > majorityCondition:
+    print ("The majority element is: ", majArray[0], "with ", majArray[1], "occurrences. ")
+else:
+    print ("The element \"", majArray[0], "\" has the most ocurrences, but it doesn't appear more than 'n/2' in the given array")
+    
+
```
Flickdm: Its not necessary technically, but its good practice  to write the code below to make the code path clearer and allows us to import functions from the script without the main running, making it easier to reuse code
```python
def main():
    pass

if __name__ == "__main__":
    main()
```

challenge_3/python/lex32pem/challenge_3.py
```diff
@@ -0,0 +1,58 @@
...
+# your program should return 7
+# 
+# Language: Python 3.6
+
+def find_Max(u):
+    dictMax = {}                            # empty dictionary
+    max = ('', 0)                           # (max element, max occurrences)
+    for n in u:                             # to apply condition on each item of array in the defined function
+        if n in dictMax:                    # if item already exist in dictionary
+            dictMax[n] += 1                 # add 1 to the value of item [n] of array
```
Flickdm: Fantastic Commenting! The idea with python is that the code can be self-documenting so using variable names that clearly describe their usage helps out a lot and makes it less code you have to write.

challenge_3/python/lex32pem/challenge_3.py
```diff
@@ -0,0 +1,58 @@
...
+# 
+# array = [2,2,3,7,5,7,7,7,4,7,2,7,4,5,6,7,7,8,6,7,7,8,10,12,29,30,19,10,7,7,7,7,7,7,7,7,7]
+# 
+# your program should return 7
+# 
+# Language: Python 3.6
+
+def find_Max(u):
+    dictMax = {}                            # empty dictionary
+    max = ('', 0)                           # (max element, max occurrences)
```
Flickdm: careful with this one, I think its stupid but python will allow you to use a variable name that is also a function which can lead to unclear and sometimes unpredictable code

challenge_3/python/lex32pem/challenge_3.py
```diff
@@ -0,0 +1,58 @@
...
+userArray = []                                                          
+majorityCondition = numberofItems/2                                     
+
+if numberofItems != 0:                                                  
+    i = 0
+    while i < numberofItems:
+        userValue =  (input ("Enter a new value to the array: "))
+        userArray.append(userValue)
+        i += 1
+    print ("Here is your Array: " , userArray)
```
Flickdm: This is just one instance but in python we have string formatters which we use to format strings.

This is just an example
```python
print("Here is your Array: {}".format(userArray))
````

File: challenge_3/python/lex32pem/challenge_3.py [hunk 5]
```diff
@@ -0,0 +1,58 @@
...
+
+majArray = find_Max(userArray)
+
+if majArray[1] > majorityCondition:
+    print ("The majority element is: ", majArray[0], "with ", majArray[1], "occurrences. ")
```
Flickdm: This is the print example again
```Python
print("The majority element is: {} with {} occurences.".format(majArray[0], majArray[1])
```

challenge_3/python/lex32pem/challenge_3.py
```diff
@@ -0,0 +1,58 @@
...
+print ("Welcome to \"Majority Element\"")                               
+print ("\"Majority Element\" is about find the element that appears more than 'n/2' in any given array of size 'n'")    
+numberofItems = int(input ("Please enter the lenght of your array: "))  
+userArray = []                                                          
+majorityCondition = numberofItems/2                                     
+
+if numberofItems != 0:                                                  
+    i = 0
+    while i < numberofItems:
+        userValue =  (input ("Enter a new value to the array: "))
```
moonpatroller: Extra parentheses  -25 points  D:

MJUIUC:
Good Stuff! Thanks @Flickdm and @moonpatroller for the input here.

PR #171, opened 2017-01-04
MajorityElement-Easier.py

drishabh:
Used "Counter" from collections module


MJUIUC:
Please Follow the guid lines for submission. Great Solution, but you can't write over someone else's files.

drishabh:
Oh sorry. It was my first time using GitHub. I will will do it correctly
now.
- Rishabh

PR #174, opened 2017-01-04
[Go] Challenge 0-5 (Unreviewed)

natdm:
Committing challenges 0 to 5

### DIFF
--- a/challenge_5/go/natdm/diff.go
+++ b/challenge_5/go/natdm/diff.go
@@ -1,0 +1,17 @@
+package natdm
+
+func diff(a, b string) (s string) {
+
+	for _, x := range []byte(b) {
+		for i, y := range []byte(a) {
+			if x == y {
+				break
+			}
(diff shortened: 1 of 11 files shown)


challenge_5/go/natdm/diff.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+		pickedIndexes = append(pickedIndexes, r)
+		out[i] = orig[r]
+	}
+	return
+}
+
+func diff(a, b string) (s string) {
+	for _, x := range []byte(b) {
+		for i, y := range []byte(a) {
```
moonpatroller: This still fails for "a" and "aa", I think.

natdm: Fixed the logic to make a map of chars (byte form) and count for each char in the original. It then loops the new word and each time it finds a character in the map with length > 0, it decrements. If the character is not there or the count is 0, it is an extra and will return it.

PR #176, opened 2017-01-04
changllenge_1-4 in python

shaowenliu:


challenge_1/python/shaowen/reverse.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Wed Jan  4 10:36:21 2017
+
+@author: shaowen Liu
+"""
+def reverse(s):
+    return ''.join(list(s)[::-1])
```
hkl0902: Why not just return `s[::-1]`?

hkl0902:
Looks good overall! The use of `dict.get(i, 0)` was pretty clever and wasn't something I knew you could do.

challenge_1/python/shaowen/reverse.py
shaowenliu: i did not know [::-1] also works for string..... Tks for the tip!

shaowenliu:
file changed 135....disaster?
it seems i fail to sync master branch properly?? 
I want to withdraw the second commit, but don't know how..
please cancel the request, and I will see what I could do.... 
sorry, still newbie to git

hkl0902:
Try merging the repo with yours. 

So try `git pull origin master`. This should result in merge conflicts in your code but you should be able to fix those. Look for the bits with `======`, `>>>>>`, and `<<<<<<` and figure out which parts are the ones you want to keep. Erase the parts you don't. Then try again.

ManuelMeraz:
Closed upon request

PR #181, opened 2017-01-04
[Haskell] Challenge 2 (ReadyForMerge)

halogenandtoast:


Flickdm:
Overall code was clear and concise. My knowledge of Haskell syntax is limited but the logic was clear and so was the style making it easy to read.

challenge_2/haskell/halogenandtoast/src/Challenge.hs
```diff
@@ -0,0 +1,21 @@
...
+
+import Data.Maybe ( fromMaybe )
+import Data.HashMap.Strict ( alter
+                           , toList
+                           , filterWithKey
+                           , empty
+                           , HashMap(..)
+                           )
+
+counts :: [String] -> HashMap String Int
```
erocs: I would alter this to be generically applicable.
counts :: (Eq a, Hashable a) => [a] -> HashMap a Int

halogenandtoast: Good point, I did this in challenge 3 with this function.

halogenandtoast:
Feedback addressed.

PR #186, opened 2017-01-04
challenge_0 in python

returnlove:


returnlove:
@ningyuansg whats the next step? looks like its not merged yet. I'm trying to create another pull request for challenge 2 but not able get through after clicking on 'new pull request'

ningyuansg:
@returnlove Hello, hello!

What I did was 'approve' your pull request. What that means, essentially, is that I reviewed your challenge #0 hello world code, and thought to myself, 'yep, that works'. 

What happens next is someone from the administrative will pop by and **merge** your pull request. This normally only happens after they have deemed that enough people (mostly fellow participants like me) have reviewed your code.

You say you are trying to create another pull request for challenge 2. This doesn't work for you because pull requests work by **branches**, not **commits**. So if you want to submit other challenges through separate pull requests (which is as of now the preferred method of submitting solutions), you will have to move your challenge 2 solution to a separate branch. 

What have have already done, instead, is commit challenges 1 & 2 to the same initial branch that you have committed challenge 0 (which is the same branch corresponding to this pull request). Since, as previously mentioned, pull requests works by branches and not commits, your solutions for 1 and 2 have been automatically added to this pull request.

Having said that, it is better to stick with this for now (multiple challenges in one pull request) than to change it (one pull request per challenge), since that process requires quite a bit of git magic. Instead, go to the edit button next to the title and add in challenge 1 & 2. Wait for more reviewers, then wait for a admin to merge.

jarydremillard:
bypass

PR #187, opened 2017-01-04
[Python] Challenge #5 (Unreviewed)

ningyuansg:
This solution uses the in-built `Counter` object from the `collections` module. 

We count the identity and occurence of letters in the strings s and t, then compare them. The odd letter out of the two is returned.


hkl0902:
Looks good. 

Style-wise though, why did you put the function within a class? Why not simply make it a global function?

PR #188, opened 2017-01-04
Suggestion for unit test (python) in challenge #5

ningyuansg:
This adds parentheses to the print statements so the unit test written by @MJUIUC can run on both python 2 and 3.

### DIFF
--- a/challenge_5/python/mjuiuc/test.py
+++ b/challenge_5/python/mjuiuc/test.py
@@ -12,7 +12,7 @@ class OutcomesTest(unittest.TestCase):
 		t1 = time.time()
 		case1out = case1in.findTheDifference("abcd","abcde")
 		t2 = time.time()
-		print 'Runtime of Case1: ' + str(t2-t1) + ' seconds' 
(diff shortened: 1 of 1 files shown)


challenge_5/python/mjuiuc/test.py
```diff
@@ -47,8 +47,8 @@ def test_case5(self):
 		t1 = time.time()
 		case5out = case5in.findTheDifference( finalTestStringS, finalTestStringT + 'p')
 		t2 = time.time()
-		print 'Runtime of Case5: ' + str(t2-t1) + ' seconds' 
+		print('Runtime of Case5: ' + str(t2-t1) + ' seconds')
 		self.assertEqual( case5out, 'p')
 
 if __name__ == '__main__':
-	unittest.main()
```
ningyuansg: It seems the difference between these two lines is a whitespace at the end? This was not an intended change, but I don't think it matters either way.

MJUIUC:
Thanks guys. I'll rework the print statements. I know I could just merge this, but I added another test case and I don't want any merge conflicts.

PR #195, opened 2017-01-04
[x86] challenge_1 [under review]

x95:


x95:
but where would be the challenge ;)
Also I can create the data structures in C and use them in asm.

challenge_1/x86/x95/src/reversestring.s
```diff
@@ -0,0 +1,90 @@
...
+    syscall
+
+#Function for reversing an input string. Stack contains inputstring as well as a buffer. The caller needs to provide the buffer.
+reverse_string:
+	#x86 Function Prolog
+	pushq %rbp
+	movq %rsp, %rbp
+	#save registers before smearing them
+	pushq %rsi
+	pushq %rdi
```
halogenandtoast: Do you have the order of rsi and rdi swapped here for some reason? rdi is supposed to be argument 1 and rsi argument 2.

I know very little abut x86 assembly so I might be wrong about this.

x95: The order of saving registers doesn't matter. It is just to preserve the registers of the caller (if it used them). However there is an error in the epilog which I will fix.

jarydremillard:
I'll approve.

PR #196, opened 2017-01-04
Adding Solutions to Challenge 5

Isaiah-Simon:


challenge_5/Java/isaiah/Test.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    // returns a random integer between 0 and n-1
+    public static int uniform(int n) {
+        return (int) (Math.random() * n);
+    }
+
+    // take as input an array of strings and rearrange them in random order
+    public static String shuffle(char[] a) {
+        int n = a.length;
```
moonpatroller: I would just use Collections.shuffle.

PR #198, opened 2017-01-04
[Python 3] Challenge 3

OkraArmband:


hkl0902:
The logic seems sound. 

Why did you say the same thing in your README as you did in the comments in the code? Things like the definition of majority element probably shouldn't be in the code while it definitely should be in the README. 

Overall though, looks good.

OkraArmband:
I've never _truly_ thoroughly documented my code (or had a README), so I wasn't sure about it all, I guess. Is there a rule or convention about what goes where? Is the code comment meant to be more functional, and the README more explanatory?

Thanks for the input!

hkl0902:
No problem!

It generally doesn't matter. For large projects, you would want to put things like `majority element` in the code as the README there should explain the project as a whole instead of the implementation in the code. For small ones (like this), if it's in the README, it doesn't need to be in the code.

And this is more of a nitpick (and really, it's just a style preference) but I prefer to put comments about methods/functions outside the method/function. This way, if I fold the code, the comments don't disappear with it. I also like to capitalize the parameters, so `LST` instead of `lst`. Finally, don't separate what you're returning into two separate paragraphs with a paragraph in between. `return: majority element. None if no majority element is found` would be easier to read.

PR #203, opened 2017-01-04
[Python] Challenge 4 and 5 (Unreviewed)

Atakyn:


hkl0902:
Looks good although Solution 4 is a little odd. Look at my solution to challenge 4 (I did it in python) for a more concise answer.

Atakyn:
@hkl0902 Interesting approach and though more concise, I'll stick with what I have currently

PR #204, opened 2017-01-04
Python Challenges #0-5

Loran425:


challenge_3/python/Loran425/src/main.py
```diff
@@ -0,0 +1,22 @@
+__author__ = "Loran425"
+
+
+def majority(data):
+    count = {}
+    output = None  # Have a value to return if there is no majority
+    for char in data:
+        # Dirty dirty dirty
```
hkl0902: try using [`dictdefault`](https://docs.python.org/2/library/collections.html#collections.defaultdict)

challenge_3/python/Loran425/src/main.py
```diff
@@ -0,0 +1,22 @@
...
+        try:
+            count[char] += 1
+        except KeyError:
+            count[char] = 1
+    for x in count.keys():
+        if count[x] >= len(count) // 2:
```
hkl0902: You can do this check in the first for loop above and let you end the function faster.

challenge_3/python/Loran425/src/main.py
```diff
@@ -0,0 +1,22 @@
...
+            output = x
```
hkl0902: Rather than simply setting output, simply return `x`. If `x` is at the front of `count.keys()`, why check the rest?

challenge_5/python/Loran425/src/FindTheDifference.py
```diff
@@ -0,0 +1,23 @@
+__author__ = "Loran425"
+
+
+class Solution(object):
+    def findTheDifference(self, s, t):
+        count = {}
```
hkl0902: There's a [Counter](https://docs.python.org/2/library/collections.html#collections.Counter) class for this. May be useful in the future.

challenge_3/python/Loran425/src/main.py
Loran425: Ah yes and then if i reach majority i can ignore the remainder of the data. Implemented in updates

challenge_3/python/Loran425/src/main.py
Loran425: Updated Implementation from previous comments fixes this problem also.

challenge_5/python/Loran425/src/FindTheDifference.py
Loran425: Huh that's pretty cool. Lots of fun toys in the collections library.

MJUIUC:
I think these solutions are great! Nice job on everything.

PR #208, opened 2017-01-05
[Python] Challenge 3

psonghi:


hkl0902:
Challenge 3 looks good. However, you might want to consider using the [Counter](https://docs.python.org/dev/library/collections.html#collections.Counter) class. The way you're doing it counts how many of that element there is more than once.

psonghi:
Thanks! I looked up [Counter](https://pymotw.com/2/collections/counter.html), and it does indeed look useful. 👍

challenge_3/python/psonghi/src/majority_element.py
```diff
@@ -0,0 +1,9 @@
+def majority_element(array):
+    half_time = len(array)/2
+    for element in array:
+        if array.count(element) > half_time:
+            return element
+
+# testing majority_element
+array = [2,2,3,7,5,7,7,7,4,7,2,7,4,5,6,7,7,8,6,7,7,8,10,12,29,30,19,10,7,7,7,7,7,7,7,7,7]
+print(majority_element(array))
```
Loran425: I would recommend you move the array declaration and print call into a main so that you could import this file later. It doesn't make a huge difference for this but its a good habit to get into. 

````python
if __name__ == "__main__": 
    array = [2,2,3,7,5,7,7,7,4,7,2,7,4,5,6,7,7,8,6,7,7,8,10,12,29,30,19,10,7,7,7,7,7,7,7,7,7]
    print(majority_element(array))
````

File: challenge_3/python/psonghi/src/majority_element.py [hunk 2]
```diff
@@ -0,0 +1,9 @@
+def majority_element(array):
+    half_time = len(array)/2
+    for element in array:
+        if array.count(element) > half_time:
```
Loran425: Remember that one of the goals of this challenge is to implement it in O(n) complexity. 

With the counter class mentioned previously you should have no problem reducing this from O(n^2) to O(n)

Good Luck

PR #213, opened 2017-01-05
[Python] Challenge 1 (Unreviewed)

thomas-vanderwal:


challenge_2/python/tvanderwal/src/challenge_2.py
```diff
@@ -0,0 +1,7 @@
+from collections import Counter
+#extra credit list
+# Added 5 to the end so the expected output in instructions would be correct
+myList = (2,'a','l',3,'l',4,'k',2,3,4,'a',6,'c',4,'m',6,'m','k',9,10,9,8,7,8,10,7,5)
+
+#use counter to see if item is repeated and a list comprehension to store them
+print([k for k, v in Counter(myList).items() if v == 1])
```
Loran425: Very clean, Normally I'd say wrapping it in a function for import to other cases would be good but since you've got a one liner for it no real need in my mind. Fantastic work!

challenge_3/python/tvanderwal/src/challenge_3.py
```diff
@@ -0,0 +1,9 @@
+#Get counter to count the items in our list
+from collections import Counter
+
+#given list for challenge
+myList = (2,2,3,7,5,7,7,7,4,7,2,7,4,5,6,7,7,8,6,7,7,8,10,12,29,30,19,10,7,7,7,7,7,7,7,7,7)
+
+#Use a list comprehension and only grab an item if it makes up more than half the items
+print([k for k,v in Counter(myList).items() if v > len(myList)/2])
```
Loran425: This one could be worth breaking out for very long lists as you miss the opportunity for early termination in the case that any count exceeds the majority. (Not needed for this case just keep it in mind for large data sets or frequently called functions)

thomas-vanderwal:
Added challenge 4

challenge_3/python/tvanderwal/src/challenge_3.py
thomas-vanderwal: Thanks for the feedback. You think .most_common() would be better? I believe it does the same as items() however sorts from greatest number of occurrences to least. Could probably just check the first nested list and see if it's greater than n/2

Loran425: I _think_ yours is actually more efficient as it doesn't sort the array just iterates over it.

challenge_2/python/tvanderwal/src/challenge_2.py
OkraArmband: I think that it's inefficient to have a loop call Counter each time, since Counter creates a dictionary (hash table) each time you call it. It might be better to first create a variable in which you store the Counter information, and then loop over that once, so you have a "looping instance" occur twice, rather than a looping instance occur within your `for` loop. I may be wrong, but I believe this is O(n^2) rather than O(n).

That said, it is correct! Very nicely styled.

challenge_3/python/tvanderwal/src/challenge_3.py
OkraArmband: Same comments from the previous challenge apply here. Still good style. Possibly inefficient.

PR #216, opened 2017-01-05
Challenge 2 Python

briguy-official:


challenge_3/c/boilerup_brian/src/majority.c
```diff
@@ -0,0 +1,60 @@
...
+}
+
+int getMode(int set[], int size){
+    int i;
+    int mode_index = 0;
+    int count = 1;
+    for(i = 1; i < size; i++){
+        if(set[i] == set[mode_index]){
+            count ++;
+        }
```
jarydremillard: Its a minor detail, but from what I have seen many professional programmers do is simply do `} else {` instead of ```
}
else {
```

Which seems odd because you can save some lines by doing the first format. Is there a reason why you dont?

Comment (briguy-official) [inline challenge_3/c/boilerup_brian/src/majority.c]: I just like the way 
```
{
else{
```
looks just a bit better in terms of readability.

jarydremillard:
Hi @briguy-official Please sign our CLA so we can merge your code. If you have any questions, please reach out to an administrator or moderator in the Slack channel. Thank you.

PR #218, opened 2017-01-05
[C++] Challenge 1 (Pending)

QQChowey:
Solution for Reversing a String

### DIFF
--- a/challenge_1/cpp/chowey/main.cpp
+++ b/challenge_1/cpp/chowey/main.cpp
@@ -4,7 +4,7 @@
 int main() {
     std::string input;
     std::cout << "Enter a string to be reversed:" << std::endl;
-    std::cin >> input;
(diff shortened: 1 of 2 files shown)


halosman:
Solution is good but I have a few comments:
- You are calling ```input.length()``` in the for loop condition and when printing the string to the screen. Since you are calling it twice, it would be better to use a constant since this value will not change. Its true that in this case it does not make much difference, but I think it's a good habit to get into.
- There is no input dialog for the user. It would be good to have a message or something just so the user knows the program is working.

QQChowey:
Alright thank you, added changes to reflect your comments.

QQChowey:
My apologies, oops. Was just doing a few quick edits without actually running the program to test for bugs. Definitely a bad thing to do. Changed cin to getline; thanks for feedback!

PR #220, opened 2017-01-05
Challenge 3 Complete in C++(Pending) 

BlakeBarnes00:
Completed Challenge 3 in C++ using Vectors.

I do have to admit that I used manuel's way of counting as it was more efficient due to it being with pointers. So credit to him for that part, everything else I have recently been learning myself.


challenge_3/cpp/blakebarnes00/src/main.cpp
```diff
@@ -0,0 +1,54 @@
...
+   */
+  int max = array.back();
+
+  /*
+   * The idea behind using the counter like this is so that we can use it in
+   * more than one place. 
+   *
+   * We allocate (in this case) 30 spaces + 1 buffer to prevent overflow.
+   */
+  int* counter = new int[max + 1]();
```
erocs: Use `std::vector<int> counter(max + 1);` and remove the delete statement below. There's almost no reason to manage explicit allocation with new / delete when using the stl and leveraging it to handle memory makes everything safer.

challenge_3/cpp/blakebarnes00/src/main.cpp
```diff
@@ -0,0 +1,54 @@
+#include <iostream>
+#include <vector>
+#include <algorithm>
+
+void print_array(std::vector<int> array)
```
erocs: Receive a reference instead of creating a local copy of the vector.  `std::vector<int> &array`

challenge_3/cpp/blakebarnes00/src/main.cpp
```diff
@@ -0,0 +1,54 @@
...
+{
+  for(int i = 0; i < array.size(); i++)
```
erocs: Leverage C++11's range based for loop.

```
for (auto i : array) {
  std::cout << i << " ";
}
```

challenge_3/cpp/blakebarnes00/src/main.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+  /*
+   * The idea behind using the counter like this is so that we can use it in
+   * more than one place. 
+   *
+   * We allocate (in this case) 30 spaces + 1 buffer to prevent overflow.
+   */
+  int* counter = new int[max + 1]();
+
+  for(int i = 0; i < array.size(); i++)
```
erocs: Once again, ranged based for loops are great.  :)

```
for (auto i : array) {
  counter[i]++;
}
```

challenge_3/cpp/blakebarnes00/src/main.cpp
```diff
@@ -0,0 +1,54 @@
...
+  {
+    counter[array[i]]++;
+  }
+
+  int majority = 0;
+  while(counter[majority] <= array.size()/2)
```
erocs: This loop really should be structured differently. In this challenge, you're guaranteed to have a majority number but real world users don't conform. Iterate over the array and search for the majority element instead so you won't access outside the boundaries of the container.

```
int majority = -1;
for (int i = 0; i < counter.size(); ++i) {
  if (counter[i] > array.size() / 2) {
    majority = i;
    break;
  }
}
if (majority < 0) {
  std::cerr << "Unable to find majority number." << std::endl;
  return 0;
}
std::cout << "Majority: " << majority << std::endl;
```

erocs:
I should add, they might want you to squash all of these 10 patches down into a single patch. I'm unsure how strict they will be about superfluous patches.

BlakeBarnes00:
@erocs how do I do that? They haven't seemed a bit bothered by it, but it would be good to know anyways

BlakeBarnes00:
@jacktraf It should work with any range of numbers as long as you put it in the vector array, even negatives to my knowledge.

challenge_3/cpp/blakebarnes00/src/main.cpp
BlakeBarnes00: I haven't learned about pointers much at all yet, only did what I knew.

challenge_3/cpp/blakebarnes00/src/main.cpp
BlakeBarnes00: Can you explain how that loop works? I have seen it more recently and I can only presume that it removes the middle part of the for loop?

If that is what it does, how would you do it if you are doing i-- instead?

challenge_3/cpp/blakebarnes00/src/main.cpp
erocs: No worries! A reference is a... nicer pointer? The variable access would work exactly the same way in your function so no other code would need to be changed. The compiler would just be utilizing a pointer to the vector under the covers instead of making a local copy every time this gets called.

challenge_3/cpp/blakebarnes00/src/main.cpp
erocs: Think of it this way. Since this is iterating over a stl container, the compiler transforms the ranged based for loop into 
```
for (std::vector<int>::iterator it = array.begin(); it != array.end(); it++) {
  int i = *it;
```

If you need to iterate in reverse, you would need to use a traditional loop. For vector you could use a reverse iterator, instead of the forward iterator in the loop I have above. You could call std::reverse on the container to reverse it and then just loop forward over it. Or you could use the traditional index based approach.

erocs:
So the input array is sorted, the maximum value is found, and that is used to allocate the counting array. Let's say it is given an array containing [-100, 1]. The maximum number is 1 so it will allocate 1 int for the counting array, already creating a problem. Then while gathering the counts, it will address the element at -100 from the start of the array to increment its count. This location is outside of the array bounds so it will overwrite arbitrary memory either crashing or causing a security vulnerability.

Let's look at an array of [0, 2147483647], the latter value being the maximum value for a signed 32-bit integer. The code will end up allocating 2Gb of memory for the counting array. Let's assume this succeeds, it will then initialize that memory to make every element 0, which will take a while. The counts will then be calculated, and finally the code will iterate over the entire counting array, taking even longer, to find the majority number. So, if the allocation is successful it will eventually work.

Other posted solutions (likely in other languages) have used maps / hash tables to store the value -> count mappings. The C++ equivalent is std::map.

challenge_3/cpp/blakebarnes00/src/main.cpp
ManuelMeraz: Wow this is good information.

challenge_3/cpp/blakebarnes00/src/main.cpp
BlakeBarnes00: Where else in my program would I use that other than in the print_array function?

challenge_3/cpp/blakebarnes00/src/main.cpp
BlakeBarnes00: So I would do this for reverse? 

```c++
void print_array(std::vector<int> &array)
{
    std::reverse(array.begin(), array.end());
    for(auto i : array)
    {
        std::cout << i << " ";
    }
    std::cout << std::endl;
}
```

challenge_3/cpp/blakebarnes00/src/main.cpp
BlakeBarnes00: Can you explain what this is doing to the vector that was just created? (Sorry for all the questions, just really would like explanations so I will remember them.)

challenge_3/cpp/blakebarnes00/src/main.cpp
erocs: Yes that's one way. I would prefer the reverse iterator or reverse index to not modify the parameter or use extra processing time. Here's some examples.
https://gist.github.com/erocs/eced45eb4af154da991ec64500bb46f8

challenge_3/cpp/blakebarnes00/src/main.cpp
erocs: It's a way to pass parameters by pointer without all of the special pointer syntax. Since print_array is your only method, that's the only location it's applicable.

challenge_3/cpp/blakebarnes00/src/main.cpp
erocs: Exactly the same thing your loop was doing. It's just less typing and looks a bit cleaner. It's counting how many occurrences of each integer exist in the input array.

PR #224, opened 2017-01-05
[Python] Challenge 1 (ReadyForMerge)

jarydremillard:


jacktraf:
It works, follows rules for path structure.  No real need for a readme I guess.

Also:
```
def reverse(string):
 	return string[::-1]
 
-
 if __name__ == '__main__':
 	print(reverse(sys.argv[1])
```
works just as well.  No real need for `" ".join(sys.argv[1:])`

PR #228, opened 2017-01-05
Challenge 1 python 

py2k5:


challenge_1/python/py2k5/reverseString.py
```diff
@@ -0,0 +1,3 @@
+str="Hello world"
+revStr=''.join(reversed(str))
```
jarydremillard: First of all, thank you for submitting. Second, don't use `str` as a variable name. That is a reserved keyword for Python. This can cause a lot of issues with software you build.

Lastly, instead of joining spaces, and reversing the string (which works, great job), you could save a lot more memory by doing `print(myStr[::-1])`

PR #229, opened 2017-01-05
[python] Challenge 2 (ReadyToMerge)

mindm:
First I tried to parse the input as JSON but it didn't quite work in the extra credit feature unless you escaped the quotes which seemed ugly.

### DIFF
--- a/challenge_2/python/mindm/src/single.py
+++ b/challenge_2/python/mindm/src/single.py
@@ -14,7 +14,10 @@ def main():
         exit(1)
 
     inlist = args_to_list(args)
-    test_alnum(inlist)
+    
+    if test_alnum(inlist) == False:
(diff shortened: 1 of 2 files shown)


challenge_2/python/mindm/src/single.py
```diff
@@ -0,0 +1,58 @@
...
+    return arg_list
+
+
+def test_alnum(arr):
+    """ Exits if list contains non-alphanumeric strings
+    """
+    for element in arr:
+        if not element.isalnum():
+            print("Error: '{}' is not alphanumeric.".format(element))
+            exit(1)
```
zmiller91: This function shouldn't be exiting, let the function that's calling this one make that decision.  This should return `True` or `False` then your `main()` method should exit.

PR #230, opened 2017-01-05
[Python] Challenge 3 (Pending)

mindm:
Based pretty much on the last challenge.

### DIFF
--- a/challenge_3/python/mindm/src/majority.py
+++ b/challenge_3/python/mindm/src/majority.py
@@ -4,6 +4,37 @@ import sys
 from collections import defaultdict
 
 
+def main():
+    if len(sys.argv) == 1:          # Arguments provided by piping in shell
+        args = sys.stdin.read()
+    elif len(sys.argv) == 2:        # Arguments provided as command line args
+        args = sys.argv[1]
+    else:
+        print("Error: too many arguments")
+        exit(1)
+
+    inlist = args_to_list(args)
+    test_digit(inlist)
+
+    # Map elements to a dictionary where the key is the element and increment
+    # the value (default value is 0 for each key initially)
+    sum_dict = defaultdict(int)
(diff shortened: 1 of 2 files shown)


challenge_3/python/mindm/src/majority.py
```diff
@@ -0,0 +1,52 @@
...
+def test_digit(arr):
+    """ Exits if list contains non-numeric strings
+    """
+    for element in arr:
+        if not element.isdigit():
+            print("Error: '{}' is not numeric.".format(element))
+            exit(1)
+
+
+if len(sys.argv) == 1:              # Arguments provided by piping in shell
```
jarydremillard: Its good practice to handle argument vectors in the beginning of the program instead of running half of your program then handle arguments.

challenge_3/python/mindm/src/majority.py
```diff
@@ -0,0 +1,52 @@
...
+import sys
+from collections import defaultdict
+
+
+def args_to_list(arg_string):
+    """ Parses argument-string to a list
+    """
+    # Strip whitespace -> strip brackets -> split to substrings ->
+    # -> strip whitespace
+    arg_list = [x.strip() for x in arg_string.strip().strip("[]").split(',')]
```
jarydremillard: You perform `.strip()` on `x` twice in the same line. Why?

mindm: I noticed that when piping arguments in for example from a file, there would be some extra whitespace around the arguments. The other .strip() happens after the string is split to elements so that the spaces between commas (if any) are removed.

challenge_3/python/mindm/src/majority.py
mindm: Thanks for the input, I will change this.

PR #234, opened 2017-01-05
C challenge 3 / update challenge 2

nasciiboy:
~(__^>


nasciiboy:
[C] Challenge 4

ManuelMeraz:
- [ ] **Code Correctness**
- [ ] **Code Style**
@nasciiboy Please add comments and document your algorithm. Your code is very difficult to read. 
- [ ] **Code Documentation**
- [ ] **Path Structure**
- [ ] **Merge Pull Request**

nasciiboy:
(sorry google translate) I'm from the school that does not add comments, exercises 3 and 4 only use a simple binary tree. 5 simply implement a custom strcmp

ManuelMeraz:
@nasciiboy 

Can you go into more detail? How do you use your simple binary tree and your custom strcomp? What is your algorithm in detail?

nasciiboy:
3: Enter elements into tre, count items
   Print items greather than or equal to someting (n/2)

4: Enter elements into tre
   Reverse tree nodes

5: Compare character to character
   return position of the first element different, otherwise return 0

ManuelMeraz:
@nasciiboy is there any particular reason you chose a binary tree?

nasciiboy:
@ManuelMeraz I can not think of a better solution for an array of arbitrary lengths and values

nasciiboy:
6 ==> two nested loops

ManuelMeraz:
@nasciiboy 

That method you use to solve with this binary tree is the same count you did for ch 1. You iterate through every element in your list and then for every element, you update a node in your tree. This algorithm is dependent on 2 things. 

1. The height of your tree
2. The length of your lest

This means it isO(n * h), where n is the input data and h is the height of your tree. As h and n go to infinity, these 2 values become the largest factors in your algorithm.

Have you considered implementing a hashmap? This would solve these problems in o(n)

nasciiboy:
@ManuelMeraz  I am an amateur with no mathematical foundation, i do not know the hash maps, but I suppose for problem 2 and 3 it needs an inmmense amount of memory

myrrlyn:
Hashmaps are non-trivial to correctly implement. A worst-case binary tree is a linked list, and h is governed by n. In the worst case, a binary tree operates exactly as a linked list, in O(n) time; in the best case a perfect binary tree operates in O(log n) time. Insertion is roughly O(log n) as well, whereas in a linked list it averages to O(n/2).

Binary trees are great for sorting a data stream on the fly. However, repeat data is more of a problem with them than for linked lists, so be sure that in those cases (if repetition data is required) the tree is a map (storing keys and values) rather than a set (keys only).

Get Outlook for Android<https://aka.ms/ghei36>



On Sat, Jan 14, 2017 at 8:59 AM -0700, "Manuel Meraz" <notifications@github.com<mailto:notifications@github.com>> wrote:


@nasciiboy<https://github.com/nasciiboy>

That method you use to solve with this binary tree is the same count you did for ch 1. You iterate through every element in your list and then for every element, you update a node in your tree. This algorithm is dependent on 2 things.

  1.  The height of your tree
  2.  The length of your lest

This means it isO(n * h), where n is the input data and h is the height of your tree. As h and n go to infinity, these 2 values become the largest factors in your algorithm.

Have you considered implementing a hashmap? This would solve these problems in o(n)

nasciiboy:
@ManuelMeraz  if a tree receives an ordered list, it is a bad imitation of an array. I will investigate the hash maps

nasciiboy:
@ManuelMeraz The other two solutions accepted, do not handle array of arbitrary lengths and values

myrrlyn:
If you really want to have good behavior in worst case (sorted) inputs, look into an auto balancing tree. I.E. if you receive 1 2 3, the tree is an ordered list with root 1, but on receipt of 3, it would rotate so that the root is 2, with children 1 and 3.

Get Outlook for Android<https://aka.ms/ghei36>



On Sat, Jan 14, 2017 at 9:13 AM -0700, "nasciiboy" <notifications@github.com<mailto:notifications@github.com>> wrote:


@ManuelMeraz<https://github.com/ManuelMeraz> if a tree receives an ordered list, it is a bad imitation of an array. I will investigate the hash maps

ManuelMeraz:
@myrrlyn Interesting, thanks for the input! I was thinking O(lg N) for going into a tree, but didn't know about the O(N/2) for linked lists, why is that? I didn't realize C wouldn't have a hashmap library. Maybe a counter array of length max value could work, but it wouldn't for extremely large data sets.

nasciiboy:
The exercise entails not receiving an ordered list

c/2/almyk tree which does not free memory
2/c/scottsquatch  an array of 256 elements

3/C/karanchawla Is an element-by-element path through the list

myrrlyn:
C doesn't have *any* collection types.

Linked lists have O(n) access time, but if they're sorted it averages out that 1/n accesses take time 1, 1/n take time 2, ... And 1/n take time n. Averages to (n+1)/2, or roughly n/2 to access. Still linear time, but better than crawling to the end for every single access like FIFO insertion is.

Get Outlook for Android<https://aka.ms/ghei36>



On Sat, Jan 14, 2017 at 9:24 AM -0700, "Manuel Meraz" <notifications@github.com<mailto:notifications@github.com>> wrote:


@myrrlyn<https://github.com/myrrlyn> Interesting, thanks for the input! I was thinking O(lg N) for going into a tree, but didn't know about the O(N/2) for linked lists, why is that? I didn't realize C wouldn't have a hashmap library. Maybe a counter array of length max value could work, but it wouldn't for extremely large data sets.

ManuelMeraz:
Sweet thanks

@nasciiboy element by element traversal is fine, it's just nesting traversals that causes problems. Your solutions works, but wouldn't you like to improve it? You're nesting an element by element traversal with a binary tree lookup.

nasciiboy:
@ManuelMeraz  Of course, some suggestion (specify example number)

nasciiboy:
add "[C] Challenge 7"

- Challenge 7/c/karanchawla Does not work if range does not start with 0

nasciiboy:
I accept that my code is under mit licence, I do not like the permissions that the application asks for

nasciiboy:
updated challenge 7/c function sumOfRanges, more powerful faster

PR #235, opened 2017-01-05
[x86] Challenge 7 (Pending)

x95:
This solution is on fire... maths fire

EDIT: my local git was a bit screwed up because of conflicts. that's why the additional commits are showing up. Only 2 files were actually changed (current challenge)


halogenandtoast:
- [x] **Code Correctness**
- [x] **Code Style**
- [x] **Code Documentation**
- [x] **Path Structure**
- [x] **Merge Pull Request**

erocs:
I'm curious why you used rcx as the accumulator instead of the loop counter, as it's built for. (LOOP instruction) Also you could remove the final SUB by subtracting each array element from rax instead of finding the array's sum. Looks good though!

PR #236, opened 2017-01-05
challenge 3 in python

returnlove:


challenge_2/python/returnlove/src/single_number.py
```diff
@@ -20,6 +20,8 @@ def main():
 if __name__ == "__main__":
 	main()
 
+<<<<<<< HEAD
```
hkl0902: Get rid of these things.

challenge_3/python/returnlove/src/majority_element.py
```diff
@@ -0,0 +1,26 @@
+def find_majority_element(input_list):
+	majority_condition = len(input_list)/2
+	# print(majority_condition)	
+	my_dict = {}
+	# print(input_list)
+	# convert input list into dictionary, key being the element and value is the count
+	for element in input_list:
```
hkl0902: Something cool to look at is something called `defaultdict(int)`. If you try to access something not in the dictionary, it'll automatically give that value a 0.

challenge_3/python/returnlove/src/majority_element.py
```diff
@@ -0,0 +1,26 @@
+def find_majority_element(input_list):
+	majority_condition = len(input_list)/2
```
hkl0902: Make sure to floor this value.

hkl0902:
- [ ] **Code Correctness**
- [ ] **Code Style**
- [ ] **Code Documentation**: Add a README and add some comments above functions. 
- [x ] **Path Structure**
- [ ] **Merge Pull Request**

challenge_2/python/returnlove/src/single_number.py
returnlove: @hkl0902 removed the old code which was commented earlier.

challenge_3/python/returnlove/src/majority_element.py
returnlove: @hkl0902 thanks for pointing, fixed.

challenge_3/python/returnlove/src/majority_element.py
returnlove: @hkl0902 thanks.

returnlove:
@hkl0902 thanks for the review, let me know if its good now.

challenge_3/python/returnlove/src/majority_element.py
```diff
@@ -0,0 +1,28 @@
+"""
+To find the majority elements in a list.
+	Counts number of times each element is present in the list
+	prints all the elements which are present more than half of the list
+"""
+
+import math
+
+def find_majority_element(input_list):
+	majority_condition = math.floor(len(input_list)/2)
```
hkl0902: This works but in python, you can do floor division more easily. `len(input_list)//2` has the same effect as `math.floor(len(input_list)/2)`. Whichever you prefer.

challenge_3/python/returnlove/src/majority_element.py
```diff
@@ -0,0 +1,28 @@
...
+	print(majority_condition)
```
hkl0902: ?

challenge_3/python/returnlove/src/majority_element.py
```diff
@@ -0,0 +1,28 @@
...
+	my_dict = {}
+	for element in input_list:
+		if element not in my_dict.keys():
+			my_dict[element] = 1
+		else:
+			my_dict[element] += 1
+
+	for k in my_dict.keys():
+		if my_dict[k] > majority_condition:
+			print(k)
```
hkl0902: For future solutions, return the value, don't print it.

hkl0902:
- [ ] **Code Correctness**: Don't print the solution, return it for both challenges (and future ones). Your solution to challenge_3 shows why printing it can be problematic.
- [x] **Code Style**
- [x] **Code Documentation**
- [x] **Path Structure**
- [ ] **Merge Pull Request**

challenge_3/python/returnlove/src/majority_element.py
returnlove: @hkl0902 my bad, I forgot to remove that print statement. Re, returning the value, I didn't get your point. Even if we return a value in a function, we need to use print statement when calling that function. Nonetheless, I created a list where I store all the matching numbers and return the list. Let me know what you think. (I'm sure using of return statement inside the for loop creates a problem as only first attempt will get returned and it ignores the second match if any. At lease in JS).

hkl0902:
- [x] **Code Correctness**
- [x] **Code Style**
- [x] **Code Documentation**
- [x] **Path Structure**
- [x] **Merge Pull Request**

challenge_3/python/returnlove/src/majority_element.py
hkl0902: @returnlove If you did return it, you're right, it would mess up your solution. However, part of the problem is to get around that. For this one, it's not too big a problem. However, for future problems, I think it will important to return, rather than print, the answer. It's not to avoid the use of print, but rather it's to be able to unit test. If you're printing, rather than returning, to test whether or not your function works, you have to visually inspect it. If you're unit testing, you can automate that which is important for larger/more complicated problems.

returnlove: @hkl0902 got that, it makes easy to unit test. Here is my summary. So basically, you were referring to avoid printing the result directly in the function but we can use the print statement when calling the function _(print(do_something())_, because we need to see the result on console.

hkl0902: @returnlove Yup.

PR #237, opened 2017-01-05
challenge_6 and challenge_7 in scala

ndsmith3:


challenge_6/scala/ndsmith3/Ranges.scala
```diff
@@ -0,0 +1,34 @@
+import scala.annotation.tailrec
+
+object Ranges {
+  @tailrec
+  def makeRangeString(answer: List[String])(curr: Array[Int]): List[String] = {
+    val answerPlusNext       = (curr.head.toString :: answer)
+    lazy val nextIsIncrement = curr.head + 1 != curr(1)
+    lazy val lastIsDecrement = answer.head.toInt + 1 == curr.head
+    curr.length match {
+      case 1                      => if (lastIsDecrement) answerPlusNext.reverse else answer.reverse
```
erocs: Instead of the nested if I would use two guarded cases like you used for the other 2 cases below to be consistent.

challenge_6/scala/ndsmith3/Ranges.scala
```diff
@@ -0,0 +1,34 @@
+import scala.annotation.tailrec
+
+object Ranges {
+  @tailrec
+  def makeRangeString(answer: List[String])(curr: Array[Int]): List[String] = {
+    val answerPlusNext       = (curr.head.toString :: answer)
+    lazy val nextIsIncrement = curr.head + 1 != curr(1)
```
erocs: Not sure the lazy variables here really buy you much, but their overhead should be negligible anyway.

challenge_7/scala/ndsmith3/MissingNum.scala
```diff
@@ -0,0 +1,15 @@
+import scala.annotation.tailrec
+
+object MissingNum {
+  def missingNum(nums: Array[Int]): Int = {
+    @tailrec
+    def test(x: Int): Int = if (!nums.contains(x)) x else test(x + 1)
```
erocs: This doesn't meet the challenges stated constraint of O(N) time. The recursion is equivalent to a O(N) loop across the array and the contains array lookup is additional overhead beyond O(N).

challenge_6/scala/ndsmith3/Ranges.scala
ndsmith3: The lazy variables are there to prevent an exception being thrown in the case of curr or answer being empty.

challenge_6/scala/ndsmith3/Ranges.scala
ndsmith3: Changed

challenge_7/scala/ndsmith3/MissingNum.scala
ndsmith3: Fixed

erocs:
Awesome, thanks for the changes!

PR #240, opened 2017-01-05
[Python 3] Challenge_4 (Pending)

OkraArmband:
I tried to pay particular attention to my documentation here, since I did something a little extra with my code. Any comments about the documentation would be greatly appreciated!


challenge_4/python/OkraArmband/src/challenge_4.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        Swap left and right children of input BinaryTree for all subsequent children as well (recursively).
+        """
+        temp = self.left
+        self.left = self.right
+        self.right = temp
+
+
+        # Continue the recursion as long as we haven't reached the end of a branch
+        if self.left is not None:
```
hkl0902: This is more of a syntaxtic thing but you could just say `if self.left`. That's a check for whether or not `self.left` is None or not.

challenge_4/python/OkraArmband/src/challenge_4.py
```diff
@@ -0,0 +1,49 @@
...
+            self.left.invertTree()
+        if self.right is not None:
+            self.right.invertTree()
+
+    @classmethod
+    def from_list(cls,data):
```
hkl0902: It's clear that you've set a precondition for the user and while it's unlikely that someone like yourself would break that precondition, it is possible for someone to pass in `[2, 3]` in which case `root = cls(data)` would make `root` an array.

hkl0902:
Solution looks good.

challenge_4/python/OkraArmband/src/challenge_4.py
OkraArmband: Yeah, I wanted to make this generalizable so that you could use any type of data type in the sort of structure where you have a "main" thing connected to two children. BSTs limit this to numbers, so I guess the thing to do if you're trying to enforce that is just throw an error message in the constructor if they pass in a list, right?

hkl0902: If they pass in anything other than a number, yes. After all, it's possible for them to pass in a completely different object (say a dictionary, or  just another BST). If you did want to do the extra work, it'd be easier to check if what they've pass in is a number instead of checking for all the things that they shouldn't have passed in. For the purposes of this challenge, I don't think it's necessary.

PR #242, opened 2017-01-05
Challenge 4 and 7!

wOstensen:


challenge_7/python/wost/find_missing_number.py
```diff
@@ -0,0 +1,14 @@
+def find_biggest(a):
+	biggest = a[0]
+	for i in a:
+		if i >= biggest:
+			biggest = i
+	return biggest
+
+def find_missing_number(c):
+	b = find_biggest(c)
```
zmiller91: will this work if your range does not start at zero? For example, what if the range was `[9, 7, 5, 8]` would your solution return 6?

wOstensen: Hm, no. I'd have to find the smallest number as well.
I'll update this when I find out about this! :)

wOstensen: NOW it should work :)

PR #243, opened 2017-01-05
Challenge #6 and #7 in Java (ReadyToMerge)

zmiller91:


challenge_7/java/zmiller91/MissingNumber.java
```diff
@@ -0,0 +1,31 @@
...
+        if (input.length == 0) {
+            return -1;
+        }
+        
+        // Find the sum, min, and max values
+        int max = Integer.MIN_VALUE;
+        int actualSum = 0;
+        for(int i : input) {
+            actualSum += i;
+            max = Math.max(max, i);
```
hkl0902: What if input was [0, 1, 2, 3]? Would your function still work?

zmiller91: How would I know that 4 is missing? The directions were a bit unclear. To me, it appears that sequence is correct. To be more clear, I think you should say the `N` is defined as the length of the list. Leaving `N` ambiguous makes me assume it can be anything from 0 to N where N is the largest number in the list.

I'll correct the solution, though.

zmiller91: Solution is fixed and README.md is added.  Those READMEs seem redundant and unnecessary.

hkl0902:
- [x] **Code Correctness**
- [x] **Code Style**
- [x] **Code Documentation**
- [x] **Path Structure**
- [x] **Merge Pull Request**

Looks good.

PR #244, opened 2017-01-06
[Python]Challenge 01234567 (Pending)

ekstrah:
thank you for your comments. and I made all the changes. Cheers!


challenge_1/python/ekstrah/src/a.py
```diff
@@ -0,0 +1,5 @@
+a = input("Type anything you want to print backwards : ")
+
+for i in range(len(a)-1, -1, -1): #{
+    print(a[i], end="")
+#}
```
Loran425: This works but you'll find the more common solution is to do a string slice.
Take a look through [this stuff](https://www.digitalocean.com/community/tutorials/how-to-index-and-slice-strings-in-python-3) and see if you can find the one i'm talking about.

challenge_3/python/ekstrah/src/a.py
```diff
@@ -0,0 +1,18 @@
+def findingMajoList(aList): #{
+    bList = set(aList)
```
Loran425: Because you immediately reassign bList to a list type this value is never used.

challenge_3/python/ekstrah/src/a.py
```diff
@@ -0,0 +1,18 @@
...
+    bList = list(bList)
+    for i in range(0, len(bList)): #{
```
Loran425: A major tool in python is the ability to iterate over most objects.
So instead of iterating an index and then looking up the item list you can instead iterate over the list. 

Instead of 
````Python
for i in range(0,X):
    print(bList[i])
````

you can just do 

````python
for item in bList:
    print(item)
````
Check out this [video](https://www.youtube.com/watch?v=EnSu9hHGq5o)
if you are interested in learning more about this topic (iterators) and a common followup topic (generators)

File: challenge_3/python/ekstrah/src/a.py [hunk 4]
```diff
@@ -0,0 +1,18 @@
...
+        if aList.count(bList[i]) > len(bList): #{
```
Loran425: Because you are already iterating over the list and the .count() function loops over it again your overall complexity becomes O(n^2) take a look at the [Counter](https://docs.python.org/3/library/collections.html#collections.Counter) class and see if you can find a way to reduce it to O(n)

challenge_3/python/ekstrah/src/a.py
```diff
@@ -0,0 +1,18 @@
...
+            return bList[i]
+        #}
+    #}
+#}
+
+
+
+aList = [2,2,3,7,5,7,7,7,4,7,2,7,4,5,6,7,7,8,6,7,7,8,10,12,29,30,19,10,7,7,7,7,7,7,7,7,7]
```
Loran425: I would move this declaration into your main statement down below just so you don't have free floating variables if you ever imported this.

challenge_2/python/ekstrah/src/a.py
```diff
@@ -0,0 +1,18 @@
+def findingMajoList(aList): #{
```
Loran425: Wrong challenge

challenge_4/python/ekstrah/src/a.py
```diff
@@ -0,0 +1,71 @@
...
+    aList = [4, 2, 1, 3, 7, 9, 6]
+    root_node = node(aList[0])
+
+    for num in aList: #{
+        root_node.insert(root_node, num)
+    #}
+    root_node.traverse()
+    root_node.reverse()
+    root_node.traverse()
+#}
```
Loran425: Very well done.
Remember that for python you don't need to build setters to assign values to member variables.
Your helper/recursive functions are well thought out and accomplish their intended goals nicely.

challenge_5/python/ekstrah/README.md
```diff
@@ -0,0 +1,3 @@
+Challenge 5 with Python 3.x
+
+The problem with my code here is that the Test Case Number 4 doesn't work since the input value is "" & "". (I don't think test case number 4 doesn't fit with permise)
```
Loran425: After rereading the premise of challenge #5 I believe you are correct in that test_4 doesn't  exactly fit the premise but this is simply an edge case that you can test for and return the expected output.

challenge_5/python/ekstrah/src/FindTheDifference.py
```diff
@@ -0,0 +1,8 @@
+class Solution(object):
+    def findTheDifference(self, s, t):
```
Loran425: The [Counter](https://docs.python.org/3/library/collections.html#collections.Counter) class can serve you well here also.
If you would prefer to stay away from it for some reason you could implement your own counter a dict with the frequency of each letter for example. 

To avoid the identical string output problem you can either test to see if the strings are the same or return a default value in the case you cannot find an extra letter.

challenge_7/python/ekstrah/src/find_missing_number.py
```diff
@@ -0,0 +1,8 @@
+def find_missing_number(num: list) -> int:
+    num.sort()
```
Loran425: The problem statement mentions that you should not sort the array.

challenge_2/python/ekstrah/src/a.py
```diff
@@ -1,18 +1,17 @@
-def findingMajoList(aList): #{
-    bList = set(aList)
-    bList = list(bList)
-    for i in range(0, len(bList)): #{
-        if aList.count(bList[i]) > len(bList): #{
-            return bList[i]
+def findingSingleList(aList): #{
+    bList = list(set(aList))
+    for item in bList: #{
+        if aList.count(item) == 1: #{
```
Loran425: The `.count()` method of a list iterates over the entire list which causes your solution complexity to be O(n^2) see if you can think of a way to reduce the complexity to O(n).

challenge_3/python/ekstrah/src/a.py
```diff
@@ -1,18 +1,22 @@
...
+    for num in aList: #{
+        if count != 0: #{
+            count += 1 if candi == num else -1
         #}
+        else:
+            candi = num
+            count = 1
     #}
+    return candi if aList.count(candi) > len(aList) // 2 else default ###Checking Again
 #}
```
Loran425: I'm just going have to admit that this works even though I cant for the life of me figure out how.

ManuelMeraz:
@Loran425 Is this one pending?

Loran425:
I believe the only thing this one is waiting on is an attempt at reducing the algorithmic complexity of the solutions for problems 3 & 7.

ManuelMeraz:
@Loran425 I'll close this PR for now then.

@ekstrah Open a new PR for each submission if you'd like to submit your PR's individually :). I've documented this in the README with the newest changes.

PR #245, opened 2017-01-06

CindyB:
Some commits are for others challenges and already merged. Sorry about that. I'll try to fix it for next time.


challenge_2/python/CindyB/src/singleNumber.py
```diff
@@ -0,0 +1,8 @@
+def findSingleNumber(myArray):
+    for x in myArray:
+        count = sum(map(lambda s: s == x,myArray))
```
hkl0902: You said in your README that you wanted to solve this in O(n). This solution is definitely not O(n). In fact, it's O(n^2).

You can think of the `sum` and `map` operations as for loops because in order for python to map an array and sum the contents of the array, it has to go through the entire array. There are some tricks that python can do but in the end, you have a nested for loop here.

challenge_3/python/CindyB/src/majorityElement.py
```diff
@@ -0,0 +1,7 @@
+def findMajorityElement(myArray):
+    for x in myArray:
+        count = sum(map(lambda s: s == x,myArray))
```
hkl0902: Again, this is O(n^2). I suggest looking at the [Counter](https://docs.python.org/3/library/collections.html?highlight=counter#collections.Counter) class or [defaultdict](https://docs.python.org/3/library/collections.html?highlight=counter#collections.defaultdict)

hkl0902:
- [x] **Code Correctness**: The code works, however, I suggest adding in the optimizations I pointed at above. 
- [x] **Code Style**
- [x] **Code Documentation**
- [x] **Path Structure**
- [x] **Merge Pull Request**

challenge_2/python/CindyB/src/singleNumber.py
CindyB: You're totally right. This is indeed some nested loops and so O(n^2). Thanks for pointing this out. I'm gonna check out for Counter class or defaultdict to get an idea of the O(n) solution.

PR #249, opened 2017-01-06
[Python] Challenges 5 and 7

alexbotello:
Completed


challenge_5/python/alexbotello/FindTheDifference.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Solution:
+    def findTheDifference(self, s, t):
+
+        s = dict(Counter(s))
+        t = dict(Counter(t))
+
+        for key in t.keys():
+            if key not in s.keys():
```
hkl0902: This is O(n^2) as you have a nested for loop. Are you sure there isn't a way to avoid that? Try using [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) instead of the standard `dict`.

challenge_7/python/alexbotello/missing_num.py
```diff
@@ -0,0 +1,18 @@
+def find_missing_num(array):
+	
+	sum_of_array = 0
+	sum_of_indexes = 0
+
+	for a in array:
+		sum_of_array += a
+	for b in range(1, len(array) + 1):
```
hkl0902: This works, but for the future, it may be useful to know that `1 + 2 + ... + N = N(N+1)/2`

hkl0902:
- [x] **Code Correctness**
- [x] **Code Style**
- [] **Code Documentation**: Explain your solution in your README and add documentation to your code wherever you feel is necessary (generally everywhere). The questions now are easy enough that most won't need them but it's good to get into the habit of doing them now rather than later. 
- [x] **Path Structure**
- [ ] **Merge Pull Request**

challenge_5/python/alexbotello/FindTheDifference.py
alexbotello: Okay, yeah I see that now. I'll remove the nested loop and handle the exception instead.

PR #252, opened 2017-01-06
[Ruby] Challenge 6

SelectiveAlso:
I think this is the worst code i have ever written. [r/badcode](http://reddit.com/r/badcode)

Note: The output is a hash object. You will see 
```
{1=>4, 8=>10, 12=>14}
```
instead of
```
["1->4", "8->10", "12->14"]
```

The first one is easier for me, and a but more friendly. I'f i need to change it to the second method, just tell me.


myrrlyn:
Have you tried implementing a custom class with a `YourClass#to_a` method?

```ruby
class YourClass
  @data = {}
  # do whatever manipulation of your hash happens
  def to_a
    @data.to_a.map |k, v| # map will yield subarrays of [k, v] and these can be destructured
      "#{k}->#{v}"
    end
  end
end
```

PR #255, opened 2017-01-06
[Ruby] Challenge 7

SelectiveAlso:


halogenandtoast:
There is a trick to challenge 7 which you can solve with a 1 liner and in O(n) time. Currently using `include?` in your loop gives you O(n^2) time complexity.

SelectiveAlso:
@halogenandtoast Refactored, needs review

PR #257, opened 2017-01-06
[Haskell] Challenge 5 (Pending)

halogenandtoast:


challenge_5/haskell/halogenandtoast/src/Challenge.hs
```diff
@@ -0,0 +1,28 @@
...
+toAlphaNum = subtract (ord 'a') . ord
+
+fromAlphaNum :: Int -> Char
+fromAlphaNum = chr . (+ ord 'a')
+
+trackCharacters :: [Int] -> [Char] -> [Int]
+trackCharacters acc [] = acc
+trackCharacters acc (x:xs) = trackCharacters (updateAlpha x acc) xs
+  where
+    updateAlpha c = (element (toAlphaNum x) %~ succ)
```
erocs: Small bug that makes no difference. This should use `toAlphaNum c`.

PR #258, opened 2017-01-06
[Haskell] Challenge 6 (Unreviewed)

halogenandtoast:


challenge_6/haskell/halogenandtoast/src/Challenge.hs
```diff
@@ -0,0 +1,23 @@
...
+    matches = takeWhile (\(i, y) -> x + i == y) (zip [1..] xs)
+
+ranges :: [Int] -> [String]
+ranges [] = []
+ranges xs = ranges' xs []
+  where
+    ranges' [] = id
+    ranges' ys@(h:t) = case endOfRange ys of
+                            Just (i, z) -> ranges' (drop i t) . (++ [rangeStr h z])
+                            Nothing -> ranges' t
```
erocs: You can change this to `Nothing -> ranges' t . (++ [rangeStr h h])` and get it to print singleton ranges encountered in the array. Of course, these aren't explicitly mentioned in the spec so it's a grey area.

erocs:
- [x] **Code Correctness**
- [x] **Code Style**
- [x] **Code Documentation**
- [x] **Path Structure**
- [x] **Merge Pull Request**

challenge_6/haskell/halogenandtoast/src/Challenge.hs
halogenandtoast: I believe the program specification mentioned dropping these values since they are not part of a range which is why I did it this way.

erocs: Ah yes, the only location that is stated is within a test case.

PR #261, opened 2017-01-06
[python]challenge_1-4 (ReadyForMerge)

shaowenliu:


shaowenliu:
(pending)
This is the second time I submit, due to git problem...
These code already received one review last time.

challenge_4/python/shaowen/bi_tree_reverse.py
```diff
@@ -0,0 +1,38 @@
...
+class Node():
+    def __init__(self, data, left_child = None, right_child = None):
+        self.data = data
+        self.left = left_child
+        self.right = right_child
+    
+    def has_child(self):
+        return (self.left or self.right)
+    
+def bi_insert(root, new_node):
```
OkraArmband: You may want to specify either in documentation or through an error checking procedure that `new_node` is indeed a `Node` object. I know you yourself wouldn't make the mistake of trying to add a number rather than an object, but it's a thing to take into consideration for others!

OkraArmband:
May want to change your title, since you've got challenges 1 through 4 in here!

shaowenliu:
Thanks for the advice! That's helpful!
Comments added!

OkraArmband:
- [x] **Code Correctness**
- [x] **Code Style**
- [x] **Code Documentation**
All is clear!
- [x] **Path Structure**
- [x] **Merge Pull Request**

PR #265, opened 2017-01-06
(C#)Challenge_0-2(Pending)

MrFojo:


MJUIUC:
Thanks for taking care of those deletions! Everything looks good now that those are gone.

PR #270, opened 2017-01-06
[rust] challenge 0, 1, 2

deg4uss3r:
@deg4uss3r commits for challenges 0, 1, 2 written in Rust


challenge_0/rust/deg4uss3r/challenge_0/challenge_0
```diff
@@ -0,0 +1 @@
+./target/release/challenge_0
```
myrrlyn: Don't commit compilation results, or symlinks to them.

challenge_1/rust/deg4uss3r/challenge_1/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+pub fn string_reverse (mut in_str: std::string::String) -> std::string::String {
```
myrrlyn: `String` is implicitly imported as part of the prelude. You don't need to fully-qualify it, just like you don't need to `use std::string::String;` it.

Also, this should take an `&str` input parameter, since you don't need mutable access or full ownership.

challenge_1/rust/deg4uss3r/challenge_1/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+
+    let mut out = "".to_string();
```
myrrlyn: `String::new()` will suffice.

challenge_1/rust/deg4uss3r/challenge_1/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+    let in_string = in_str.len();
+    let mut i: usize = 0;
+
+    while i < in_string {
+        let out_temp = in_str.pop().unwrap().to_string();
+        out += &out_temp;
+        i+=1;
+    }
```
myrrlyn: That's one way to do this (and explains the mutable ownership parameter), but not the most idiomatic. Strings can coerce to double-ended iterators very easily, so `in.chars().rev().collect()` will split, tweak, and rejoin the input.

challenge_2/rust/deg4uss3r/challenge_2/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+use std::vec::Vec;
+
+pub fn find_lonely (mut input_vec: Vec<u64>) -> u64 {
```
myrrlyn: May want to make that return `Option<u64>`, just in case there isn't a lonely in your input data. Safety first!

challenge_3/rust/deg4uss3r/challenge_3/src/lib.rs
```diff
@@ -0,0 +1,20 @@
+use std::vec::Vec;
+use std::collections::HashMap;
+
+pub fn majority_finder(input_vec: Vec<u64>) -> u64 {
```
myrrlyn: This should *definitely* be `Option<u64>`, since (IIRC) the spec says to do nothing if no majority is found.

challenge_3/rust/deg4uss3r/challenge_3/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+    let mut counter = HashMap::new();
+    let leng = input_vec.len()/2;
+
+    for i in input_vec{
+        let c = counter.entry(i).or_insert(1);
+        *c+=1;
```
myrrlyn: Nice.

myrrlyn:
I left a lot of notes, but none are cause for full rejection.

PR #273, opened 2017-01-07
Challenge 2

slansford:


challenge_2/python/slansford/src/challenge_2.py
```diff
@@ -0,0 +1,8 @@
+#prints the occurence that does not repeat
+#in the given array. Uses Python 3.6.0
+
+given_array = [2,'a','l',3,'l',4,'k',2,3,4,'a',6,'c',4,'m',6,'m','k',9,10,9,8,7,8,10,7]
+
+for i in given_array:
+	if given_array.count(i) == 1:
+		print(i)
```
OkraArmband: Note that in the count() function, Python is looping through all the elements in your list. So with this loop, which is itself inside your "for" loop, you are getting O(n^2) operations in your code (for each _n_, you do a `count` which takes _n_ operations). This isn't the most efficient method, but it is technically correct!

challenge_3/python/slansford/src/challenge_3.py
```diff
@@ -0,0 +1,7 @@
+given_array = [2,2,3,7,5,7,7,7,4,7,2,7,4,5,6,7,7,8,6,7,7,8,10,12,29,30,19,10,7,7,7,7,7,7,7,7,7] 
+
+for i in given_array:
+	if given_array.count(i) > len(given_array)/2:
+		print(i)
+		break
```
OkraArmband: Same issue here as in challenge 2, with the count() inside of a loop. Can be optimized.

Also, note that `len(given_array)/2` isn't _equal_ to the floor of the length divided in half. You can either use a floor function or do integer division. Assuming that `given_array.count(i)` is an integer, though, this shouldn't matter too much.

challenge_2/python/slansford/src/challenge_2.py
slansford: Thank you for your critique. What would be a more efficient model for the algorithm?

challenge_3/python/slansford/src/challenge_3.py
slansford: So should len(given_array)/2 really be len(given_array)//2?

jacktraf:
The title of this PR doesn't follow the Review Title System listed here:
https://github.com/YearOfProgramming/2017Challenges
At the bottom of the Readme

PR #282, opened 2017-01-07
[Python] Challenge #6 (ReadyToMerge)

ningyuansg:
_From README.md,_

1. A copy of the `input_list`, called `output_elements`, is created. This is to avoid modifications to the original list while we iterate through it. The `output_elements` will hold integers that belong to a range, i.e. those in [0, 1, 2, 3, 4], but not [0, 2, 4].

2. A list named `output_list` is created. This will be the list that the function will return, with elements formatted in the required style of `'x->y'`.

3. We iterate through the list with two conditionals. The first checks if for an integer `x`, if `x-1` or `x+1` exists in the `input_list`. If `x-1` and `x+1` does not exist, this means that `x` is a 'loner' that does not belong in a range, and is thus removed from consideration as our output (`output_elements.remove(integer)`.

4. The second conditional within this first iteration checks if for an integer `x`, if both `x-1` and `x+1` exists. If true, this means that the integer `x` is contained within the range (i.e. not belonging to either ends of the range). We remove this from our `output_elements` so we can simplify `output_elements` to include only the two extremes for each range (i.e. for `'x->y'`, only contain `[x, y]`).

5. We iterate once more, this time through the `output_elements` and using `enumerate`. This final iteration serves the purpose of formatting `output_elements` into the expected presentation of `output_list`.


challenge_6/python/ning/challenge_6.py
```diff
@@ -0,0 +1,19 @@
+def ranges(input_list):
+    output_elements = list(input_list)  # copy, not assign
+    output_list = []
+
+    for integer in input_list:
+        if (integer + 1 not in input_list and
```
zmiller91: Because you are using a `list` and not a `set` this `integer not in input_list` runs in O(n) time, making your algorithm run in O(n^2).  You should be able to solve this algorithm in O(n) time.

ningyuansg:
With 15d3cbd, `for integer in ...` now uses the set `input_lookup` generated from `input_list`, as advised by @zmiller91.

challenge_6/python/ning/challenge_6.py
ningyuansg: Thanks for the suggestions, TIL about sets. This portion of the code has been updated accordingly in 15d3cbd.

ningyuansg:
(I had accidentally deleted my branch `challenge_6` while working on the README.md for challenge 7 in #294)

PR #283, opened 2017-01-07
Challenges

MaxfieldHorwood:
Challenge_6 test.py is the wrong one. That can be deleted.


challenge_7/python/maxfieldhorwood/challenge_7.py
```diff
@@ -0,0 +1,25 @@
...
+    wantedTotal = 0
+    actualTotal = 0
+    arrLen = len(arr)
+    # Get the total of ALL the numbers (even the missing one)
+    # Additionally, get the total of the ACTUAL list
+    # Return the difference
+    for i in xrange(arrLen+1):
+        wantedTotal += i
+        if i != arrLen:
+            actualTotal += num[i]
```
slandau3: I think you meant arr[i]

PR #284, opened 2017-01-07
[Julia] Challenge_3 (Unreviewed)

dtcitron:
Also have code for Python Challenge 3


slandau3:
Looks good. Don't think i missed anything

PR #285, opened 2017-01-07
challenge_1 in csharp

bdoner:


bdoner:
You are right. Sorry about that. Moved solution to right folder.

PR #290, opened 2017-01-07
C challenge 3 (UnReviewed)

halosman:


challenge_3/c/scottsquatch/src/main.c
```diff
@@ -0,0 +1,94 @@
...
+int get_majoriy_element(int*, int);
+int* read_array(char*,int);
+
+int main(int argc, char* argv[])
+{
+  // Get parameters from input
+  if (argc != NUM_ARGS)
+  {
+    // Incorrect number of parameters
+    printf("Incorrect Number of parameters, expected %d but got %d\n",
```
erocs: If instead you would like to print to stderr to indicate error messages, use `fprintf(stderr, "`.

challenge_3/c/scottsquatch/src/main.c
```diff
@@ -0,0 +1,94 @@
...
+      NUM_ARGS - 1, argc - 1);
+    printf("Usage: ./majority_element <array file> <size of array in file>\n");
```
erocs: Return a non-zero value from main to indicate an error. That way a script could differentiate success from failure. Also return zero from main on success.

challenge_3/c/scottsquatch/src/main.c
```diff
@@ -0,0 +1,94 @@
...
+  }
+  else
+  {
+    // Read array from file
+    char* fileName = argv[FILE_NAME_ARG];
```
erocs: It's nice to use const to state that these values shouldn't change.

```
const char* fileName = argv[FILE_NAME_ARG];
const int arraySize = atoi(argv[ARRAY_SIZE_ARG]);
```

challenge_3/c/scottsquatch/src/main.c
```diff
@@ -0,0 +1,94 @@
...
+    int arraySize = atoi(argv[ARRAY_SIZE_ARG]);
```
erocs: Return an error if arraySize is <= 0 before using it. 0 indicates there was an error parsing the number.

challenge_3/c/scottsquatch/src/main.c
```diff
@@ -0,0 +1,94 @@
...
+
+    // Free memory
+    free(input_array);
+  }
+}
+
+// Read array from FILE
+int* read_array(char* fileName, int arraySize)
+{
+  int* array = calloc(arraySize, sizeof(int));
```
erocs: Check that array != NULL after the allocation. If it does there was an error allocating the memory.

challenge_3/c/scottsquatch/src/main.c
```diff
@@ -0,0 +1,94 @@
...
+  FILE *input = fopen(fileName, MODE_READ);
+  int index = 0;
+
+  // Read data directly into array.
+  while (fscanf(input, "%*c %d", &array[index++]) == 1
```
erocs: You'll likely want to check if fscanf failed to parse an array element so you could return an error. I'd probably use a do/while instead. Try and keep the loop conditionals easy to read for when you have to do so later.

```
int rc;
do {
  rc = fscanf(input, "%*c %d", &array[index++]);
} while (1 == rc && index < arraySize);

if (1 != rc) {
   // some error handling
}
```

challenge_3/c/scottsquatch/src/main.c
```diff
@@ -0,0 +1,94 @@
...
+    // Output to Screen for informational purpose
+    printf("array: ");
+    print_array(input_array, arraySize);
+    printf("\n");
+
+    // Print Majority Element
+    printf("Majority Element: %d\n", get_majoriy_element(input_array, arraySize));
+
+    // Free memory
+    free(input_array);
```
erocs: Just a tip, I would get in the habit of setting the freed pointer to NULL. Sure, it doesn't mean much as it stands here, but if you ever add code below this and start using that freed pointer, at least you'll crash right away and catch your bug instead of missing a potential security vuln.

challenge_3/c/scottsquatch/src/main.c
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+// Read array from FILE
+int* read_array(char* fileName, int arraySize)
+{
+  int* array = calloc(arraySize, sizeof(int));
+  FILE *input = fopen(fileName, MODE_READ);
```
erocs: Once again, test for an error opening the file by checking to see if input is NULL.

challenge_4/c/scottsquatch/src/binary_tree.h
```diff
@@ -0,0 +1,130 @@
...
+
+/* Structs */
+struct Node
+{
+  int data;
+  struct Node* left;
+  struct Node* right;
+};
+
+struct BinaryTree
```
erocs: You can use `typedef` to give these structures a type name. This would eliminate having to type struct so often below.

`typedef struct { struct Node* root; } BinaryTree;`

And from then on you can just refer to `BinaryTree`.

challenge_4/c/scottsquatch/src/binary_tree.h
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// Recursive method call
+struct Node* add_recurse(struct Node* parentNode, int data)
+{
+  // base case
+  if (parentNode == NULL)
+  {
+    // Sweet we can create the new node now
+    parentNode = calloc(1, sizeof(struct Node));
```
erocs: NULL test for allocation failure.

challenge_4/c/scottsquatch/test/binary_tree_tests.c
```diff
@@ -0,0 +1,210 @@
...
+  /* Expect tree
+        6
+       / \
+      9   3
+     /   / \
+    10  4  2
+  */
+
+  printf("Expect output below to be 6 9 3 10 4 2\n");
+  print_tree(testTree);
```
erocs: If there was a, let's call it `char *string_tree(BinaryTree tree)`, that translated the tree into a string of the format which print_tree outputs, that could also be used here to do an automated test against the expected string.

```
char* st = string_tree(testTree);
if (st == NULL) {
  // print failure
} else {
  int rc = strcmp(st, "6 9 3 10 4 2");
  if (0 != rc) {
    // print failure
  }
  free(st);
}
```

challenge_3/c/scottsquatch/src/main.c
myrrlyn: A `main` function that lacks an explicit return on the main branch implicitly returns 0. I'm not sure if that's in the standard or just a thing most compilers do, but `return 0;` is inserted at the end of `main()` in at least GCC and Clang.

PR #293, opened 2017-01-08
[bash]challenge_1(Pending)

jacktraf:
Tried to be as descriptive as possible with my comments.  If you think I went overboard on the comments, let me know lol.  Any suggestions are greatly appreciated


jacktraf:
Thanks!

PR #298, opened 2017-01-08
[Python] Challenge #7 (Pending)

ningyuansg:
1. Our solution for challenge 7 comes in the form of a function `find_missing`, taking a paramter of list `input_list`.
2. We first create a set `input_set` from `input_list`. This is as it is faster to check if an element is in a set than in a list, in python.
3. We find the minimum and maximum of `input_list`, i.e. the range of the 'list of numbers from 0 to N-1'.
4. We iterate through this range from smallest to largest, checking for each integer if it exists in `input_set` (and therefore `input_list`). If it does not exist (i.e. is _missing_ from the `input_set`), return the integer.


ningyuansg:
@slandau3 Hi! Thanks for the review. 

My alternative to the current solution would be to calculate the sum of the given list, followed by the sum of all integers from min(given list) and max(given_list). The difference between the two sums will give us the missing number. This should work in the case where the missing number is a positive, negative, or zero value.

However, as I understand it, `for ... in ...` lookup for sets in python is within O(n) complexity. This I learnt from a review received in challenge 6 (#282). If I'm not wrong, [this article on wiki.python.org](https://wiki.python.org/moin/TimeComplexity) would confirm it.

ningyuansg:
p.s. I will test the two solutions later on with `timeit` and post the results here. I currently do not have access to my laptop (and who wants to write code on a phone through SSH). Shall we keep this PR open until then?

ningyuansg:
Guys. GUYS. 

I don't know what happened to include all these god damn commits from JANUARY 4TH, but in the meanwhile, here's the results of the timeit. I have it under my directory as `time.py`, but also posting it here in case it goes missing, or the link changes will I try to fix my mess.
```
 List with 4 elements, using for ... in set (x 1 000 000)
 1.1550223514080038
 List with 4 elements, using difference of sums (x 1 000 000)
 1.391524411772641
 List with 100 elements, using for ... in set (x 1 000 000)
 8.43574248785071
 List with 100 elements, using difference of sums (x 1 000 000)
 8.94695660741872
 List with 1000 elements, using for ... in set (x 100 000)
 8.1138781071155
 List with 1000 elements, using difference of sums (x 100 000)
 8.383110816298519
```

PR #304, opened 2017-01-08
[c++] [c#] Challenges 0, 1 & 2

svclaeys:
I've separated all solutions into their own commits

### DIFF
--- a/challenge_0/cpp/sven/README.md
+++ b/challenge_0/cpp/sven/README.md
@@ -1,0 +1,11 @@
+#Challenge_0 Solution
+
+Under Windows - using the Visual studio c++ compiler:
+run "cl.exe src\HelloWorld.cpp" and run the resulting file "HelloWorld.exe".
+
+cl can be found in C:\Program Files (x86)\<Visual Studio folder>\VC\bin
+(e.g. C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin)
+or you can open a visual studio command prompt which adds the correct paths to its environment.
+
+Under Linux - using the gcc compiler:
+g++ src\HelloWorld.cpp && a.out
--- a/challenge_0/cpp/sven/src/HelloWorld.cpp
+++ b/challenge_0/cpp/sven/src/HelloWorld.cpp
@@ -1,0 +1,7 @@
+#include <iostream>
+
+int main()
+{
+	std::cout << "Hello World!" << std::endl;
+	return 0;
+}
--- a/challenge_0/csharp/sven/README.md
+++ b/challenge_0/csharp/sven/README.md
@@ -1,0 +1,7 @@
+#Challenge_0 Solution
+
+To compile run "csc.exe src\HelloWorld.cs" and run the resulting file "HelloWorld.exe".
+
+csc can be found in <windows installation folder>\Microsoft.NET\<Framework bitness>\<.NET version> 
+(e.g. C:\Windows\Microsoft.NET\Framework64\v4.0.30319)
+or you can open a visual studio command prompt which adds the correct paths to its environment.
--- a/challenge_1/cpp/sven/src/Reverse.cpp
+++ b/challenge_1/cpp/sven/src/Reverse.cpp
@@ -1,0 +1,15 @@
+#include <iostream>
+#include <string>
+
+int main(int argc, char* argv[])
+{
+	for (int i = argc - 1; i > 0; --i)
+	{
+		std::string word = argv[i];
+		for (std::string::reverse_iterator rit = word.rbegin(); rit != word.rend(); ++rit)
+			std::cout << *rit;
+		std::cout << " ";
+	}
+	std::cout << std::endl;
+	return 0;
+}
--- a/challenge_1/csharp/sven/src/Reverse.cs
+++ b/challenge_1/csharp/sven/src/Reverse.cs
@@ -1,0 +1,15 @@
+using System;
+using System.Linq;
+
+namespace Challenge_1
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Console.WriteLine(
+                args.Aggregate("", (sentence, word) => word.Aggregate("", (w, character) => character + w) + " " + sentence)
+                );
+        }
+    }
+}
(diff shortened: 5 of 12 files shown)


myrrlyn:
We squash commits when we merge, but, that's good practice for you to do locally.

challenge_0/cpp/sven/src/HelloWorld.cpp
```diff
@@ -0,0 +1,7 @@
+#include <iostream>
+
+int main()
+{
+	std::cout << "Hello World!" << std::endl;
```
myrrlyn: I *really* appreciate your use of `std::` and not `using namespace std;`. Same goes for `std::endl`.

challenge_0/csharp/sven/README.md
```diff
@@ -0,0 +1,7 @@
+#Challenge_0 Solution
+
+To compile run "csc.exe src\HelloWorld.cs" and run the resulting file "HelloWorld.exe".
+
+csc can be found in <windows installation folder>\Microsoft.NET\<Framework bitness>\<.NET version> 
+(e.g. C:\Windows\Microsoft.NET\Framework64\v4.0.30319)
+or you can open a visual studio command prompt which adds the correct paths to its environment.
```
myrrlyn: For future reference, under OSX and Linux we can either install Mono or .NET directly. With Mono, use `mcs *.cs` to build and `mono *.exe` to run.

If you commit your `.csproj` file, the official .NET `dotnet` tool can run it as well.

challenge_0/cpp/sven/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Under Windows - using the Visual studio c++ compiler:
+run "cl.exe src\HelloWorld.cpp" and run the resulting file "HelloWorld.exe".
+
+cl can be found in C:\Program Files (x86)\<Visual Studio folder>\VC\bin
+(e.g. C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin)
+or you can open a visual studio command prompt which adds the correct paths to its environment.
+
+Under Linux - using the gcc compiler:
+g++ src\HelloWorld.cpp && a.out
```
myrrlyn: (UNIX systems use `/` as the path separator, not `\`)

challenge_1/cpp/sven/src/Reverse.cpp
```diff
@@ -0,0 +1,15 @@
+#include <iostream>
+#include <string>
+
+int main(int argc, char* argv[])
+{
+	for (int i = argc - 1; i > 0; --i)
+	{
+		std::string word = argv[i];
+		for (std::string::reverse_iterator rit = word.rbegin(); rit != word.rend(); ++rit)
```
myrrlyn: Good use of the stdlib.

challenge_1/csharp/sven/src/Reverse.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Linq;
+
+namespace Challenge_1
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Console.WriteLine(
+                args.Aggregate("", (sentence, word) => word.Aggregate("", (w, character) => character + w) + " " + sentence)
```
myrrlyn: I believe you, but, could you explain what's going on here?

challenge_2/cpp/sven/src/Single.cpp
```diff
@@ -0,0 +1,22 @@
...
+		if (map.find(el) == map.cend())
+		{
+			map[el] = true;
+		}
+		else
+		{
+			map[el] = false;
+		}
+	}
+	std::cout << std::find_if(map.cbegin(), map.cend(), [](const std::unordered_map<std::string, bool>::value_type& vt) { return vt.second; })->first << std::endl;
```
myrrlyn: I believe you here as well, but I don't speak C++ well enough to know why. Can you explain in a response?

challenge_1/csharp/sven/src/Reverse.cs
svclaeys: Aggregate is a sequence accumulator. It is a linq method which applies a given function to each element of the sequence it operates on, while carrying the result forward.

I've used two Aggregate methods, one nested into the other. The effect is that the outer most Aggregate will swap the order of the words in a sequence (hello world! => world! hello).
The inner most Aggregate uses the same logic but operates on each word separately (hello => olleh)
The combination is that a given string is reversed.

Both aggregate methods I use takes 3 parameters.
I'll explain it for the outer most Aggregate:
- the seed (the empty string ""): is the initial value that will be present in the accumulator argument "sentence"
- the accumulator "sentence"
- the element to operate on "word"
The function to perform on each element is given after the "=>" operator as a lambda:
concatenate the (reverse of a) word to the front of the sentence.

challenge_2/cpp/sven/src/Single.cpp
svclaeys: Before this statement, we've build the map that contains the inputvalues as keys of the map and whether they appear once or multiple times as bool values of the map (once = true, multiple occurrences = false).

So, the only thing that is left to be done is to go trough the map and look for the first (and single) value that is "true" and return its key (number):
find_if is a method present in the <algorithm> class which goes trough a given range and returns the first element it finds for which the given predicate is true.

I use find_if by giving the entire range of the map (map.cbegin(), map.cend()) and simply use the value of the map as predicate (as said before: the value of the map will be true if the value only appeared once in the given input).
find_if returns the correct element and that's the output that we print out on the standard output.

let's go trough it step by step:
std::find_if takes 3 arguments
- the first is the beginning of the range. I use map.cbegin(), the beginning of our map
- the second is the end of the range. map.cend()
- the last argument is the predicate that is applied on each element. 
I use a lambda as predicate.
The lambda takes a single argument: "vt" of type "const std::unordered_map<std::string, bool>::value_type&" (the type of a value in an unordered_map), and returns the associated value of the element "{ return vt.second; }"

the "->first" that comes after the find_if is to retrieve the key of the found element and print it out

svclaeys:
Thanks for the review and feedback!
I know this must take up a whole lot of your time. Very much appreciated!

myrrlyn:
Yay explanations and learning! I don't speak high-level C++ _at all_, so, this is fun for me too.

PR #305, opened 2017-01-08
[rust] challenge 1

makernaren:


challenge_1/rust/makernaren/src/string_reverser.rs
```diff
@@ -0,0 +1,34 @@
...
+	let mut char_vec = vec!['a'; n];
+	// Replace the chars to tail of the vector
+	for (i, c) in input.chars().enumerate() {
+		char_vec[n-1-i] = c
+		// char_array.push(c)
+	}
+	// Iterate, copy and collect it as a string
+	let result: String = char_vec.iter().cloned().collect();
+	return result
+	// oneliner : return input.chars().rev().collect::<String>();
```
josephpd3: Rust allows you to just leave the expression of desired output on the last line of a function/block's scope so that return is only necessary if explicitly done before the end of a function/block.

For example, instead of:
```rust
let result: String = char_vec.iter().cloned().collect();
return result
```
...you could write something like:
```rust
char_vec.iter().cloned().collect::<String>()
```

makernaren: Thanks for the feedback, old habits, you know. Will try to correct it :)

myrrlyn: Expression semantics in a language that looks like C took me a *long* time to get used to, honestly.

PR #307, opened 2017-01-08
[Python]Challenge_8(ReadyForMerge)

MJUIUC:
Finally finished challenge 8. Coming up with a way to test it was harder than coming up with the actual algorithm lol.


challenge_8/python/mjuiuc/solution.py
```diff
@@ -0,0 +1,23 @@
...
+
+def solution(head):
+	lop = dict()
+	newHead = listNode()
+	temp = head
+	temp2 = newHead
+	# copy the list first
+	while temp != None:
+		temp2.val = temp.val
+		lop[temp2.val] = temp2
```
zmiller91: This makes it such that you cannot have duplicate numbers in your list.  `LinkedLists` should allow duplicate values, otherwise you've made a `LinkedSet`

challenge_8/python/mjuiuc/solution.py
```diff
@@ -0,0 +1,23 @@
...
+		temp2.next = listNode()
```
zmiller91: Does this mean you'll have an empty node at the very end? The end of a list should be null terminated.

challenge_8/python/mjuiuc/solution.py
```diff
@@ -0,0 +1,23 @@
...
+		temp2 = temp2.next
+		temp = temp.next
+	# assign the random pointers next
+	temp3 = head
+	temp4 = newHead
+	while temp3 != None:
+		temp4.random = lop[temp3.random.val]
```
zmiller91: null pointer if `temp3` doesn't have a `random`

challenge_8/python/mjuiuc/test.py
```diff
@@ -0,0 +1,93 @@
+import unittest
+import random
+import time
+from listNode import listNode
+from solution import solution
+
+class Tests(unittest.TestCase):
```
zmiller91: I know this is just test code, but did you copy and paste `test1` to `test2` and `test3`? If you find your self pressing cntrl+c, then it's probably a good opportunity for abstraction.  Also, what do these tests do? You should have python docs that describe what the test is and expected value, ideally your test methods should reflect what you're testing.

MJUIUC: Oh yea good point. I'll probably work towards that in my next set of tests.

challenge_8/python/mjuiuc/solution.py
MJUIUC: Yea the end of the list is just an empty pointer. Sentinel nodes do exist though, and technically if there is no value at next it will return None anyways. Python is too smart to let me be stupid lol. You're right though it's not defined and it should be.

challenge_8/python/mjuiuc/solution.py
MJUIUC: Oh yea you're right. Hmmmm I wonder if python has a data structure that's similar to an unordered multi map in c++.

challenge_8/python/mjuiuc/solution.py
```diff
@@ -0,0 +1,27 @@
...
+		lop[temp] = temp2
+		temp2.next = listNode()
+		temp2 = temp2.next
+		temp = temp.next
+	temp2.next = None
+	# assign the random pointers
+	temp3 = head
+	temp4 = newHead
+	while temp3.next != None:
+		temp4.random = lop[temp3.random]
```
MJUIUC: If there is no value here, I wrote the listNode class to ensure that the value would default to None.

PR #310, opened 2017-01-09
[Python] Challenge 6 (unreviewed)

thomas-vanderwal:
Added solution for Challenge 6 with unit testing in main


challenge_1/python/tvanderwal/src/challenge_1.py
```diff
@@ -0,0 +1,6 @@
+s = input('Enter string:')
+#Use extended splice
+print(s[::-1])
+
+#one liner since instructions said to make it as short as possible
```
moonpatroller: Mixing input in here makes it hard to test.

challenge_3/python/tvanderwal/src/challenge_3.py
```diff
@@ -0,0 +1,9 @@
+#Get counter to count the items in our list
+from collections import Counter
+
+#given list for challenge
+myList = (2,2,3,7,5,7,7,7,4,7,2,7,4,5,6,7,7,8,6,7,7,8,10,12,29,30,19,10,7,7,7,7,7,7,7,7,7)
+
+#Use a list comprehension and only grab an item if it makes up more than half the items
+print([k for k,v in Counter(myList).items() if v > len(myList)/2])
```
moonpatroller: This also looks hard to test.  Python's modules should make stuff like this easy to separate and test.

challenge_5/python/tvanderwal/src/FindTheDifference.py
```diff
@@ -0,0 +1,19 @@
...
+        :return:  String
+        """
+
+        #Turn strings into sorted lists
+        sList = sorted(s)
+        tlist = sorted(t)
+
+        #enumerate over ever char in tList. If it isn't equal to the same spot in sList then it was added
+        for index,val in enumerate(tlist):
+            #If last value is the character aded
```
moonpatroller: Nice algo.  You could also do this with zip and/or a comprehension.

challenge_7/python/tvanderwal/src/findMissingNumber.py
```diff
@@ -0,0 +1,12 @@
+#Challenge 7: Return a missing integer from an unsort list of integers
+
+def findMissingNumber(intList):
+    #sort list
+    intList.sort()
```
moonpatroller: Sorting is O(nlogn).    There's a O(n) solution.  But still nice code.

thomas-vanderwal: Thanks! I actually just updated it to be O(n), but I think there is even a better algo to find the sum of the range over my for loop. Thanks for the feedback though. Sorry for the hard to test code, I was taking the README examples and just using them. Now I've been building them easier to use unittesting so hopefully that helps

challenge_5/python/tvanderwal/src/FindTheDifference.py
thomas-vanderwal: I'll have to look into zip, still pretty green to python. Comprehension would be a really good idea here! Thanks for the heads up,

challenge_7/python/tvanderwal/src/findMissingNumber.py
```diff
@@ -1,12 +1,21 @@
...
 
     for i, val in enumerate(intList):
         if i != val:
             return i
+"""
+
+def findMissingNumber(intList):
+    sumInts = sum(intList)
+    sumRange = 0
+    for i in range(1, len(intList) + 1):
```
moonpatroller: The sum from 1 to n is n * (n+1) / 2, so you don't need to loop through from 1 to n, but your method is still O(n).

moonpatroller: - [x] **Code Correctness**
- [x] **Code Style**
- [x] **Code Documentation**
- [x] **Path Structure**
- [ ] **Merge Pull Request**

PR #312, opened 2017-01-09
[Python] Challenge #7 (Approved)

ningyuansg:
This is a remake of #298, a pull request of the same challenge and solution. Unfortunately, I somehow screwed that pull request up by adding 88 commits (I still don't know how that happened).

Last review was by @slandau3 (requested changes). Since then, I've added an alternative solution using the difference between the sum of given list and the sum of all integers within the range of the min and max of the given list.

In #298 I suggest that `for ... in ...` lookup in sets for python is an O(n) operation ([reference](https://wiki.python.org/moin/TimeComplexity)). 

Then I compared the two solutions to get the following `timeit` results:

```
List with 4 elements, using for ... in set (x 1 000 000)
1.1550223514080038
List with 4 elements, using difference of sums (x 1 000 000)
1.391524411772641
List with 100 elements, using for ... in set (x 1 000 000)
8.43574248785071
List with 100 elements, using difference of sums (x 1 000 000)
8.94695660741872
List with 1000 elements, using for ... in set (x 100 000)
8.1138781071155
List with 1000 elements, using difference of sums (x 100 000)
8.383110816298519
```


challenge_7/python/ning/README.md
```diff
@@ -0,0 +1,10 @@
+# Challenge 7, Find The Missing Number
+
+1. Our solution for challenge 7 comes in the form of a function `find_missing`, taking a paramter of list `input_list`.
```
zmiller91: `find_missing` runs in O(n^2), did you mean `alt_find_missing`?

challenge_7/python/ning/challenge_7.py
```diff
@@ -0,0 +1,10 @@
+def find_missing(input_list):
```
zmiller91: is this your solution or is the other? delete the one you dont want.

ningyuansg: Fixed with c75004a, thanks!

PR #313, opened 2017-01-09
[Python] Challenge 7 (ReadyToMerge)

mindm:


karanchawla:
Good to go!

You can make your solution further compact by doing this:  `sum(xrange(a[0],a[-1]+1)) - sum(a)`
assuming `a` is the input list.

PR #314, opened 2017-01-09
[C] Challenge 5 (Unreviewed) and Challenge 8 (Pending)

karanchawla:
Catching up slowly and completing the ones I missed. Probably a really poor implementation - would love recommendations on improving this. 

Peace,
Karan.


challenge_8/C/randomDeepCopy.c
```diff
@@ -0,0 +1,111 @@
...
+		temp2->next = old;
+		temp = old;
+	}
+
+	temp = head;
+	//copy arbitrary matching
+	while(temp && temp->next)
+	{
+
+		temp->next->random = temp->random->next;
```
zmiller91: segfault when `temp->random` is `NULL`

challenge_5/C/karanchawla/src/challenge5.c
```diff
@@ -0,0 +1,109 @@
...
+    int pos = rand() % (size-1);
+    
+    for(int i=0;i<size;i++)
+    {
+        char temp;
+
+        if(i==pos)
+        {
+            temp = t[i];
+            t[i] = 'd';
```
myrrlyn: Your comment doesn't match reality :p

challenge_5/C/karanchawla/src/challenge5.c
```diff
@@ -0,0 +1,109 @@
...
+    
+    char extraChar[1];
+    int flag = 0;
+
+    //find the added character
+    for(int i=0; i<size-1; i++)
+    {
+        if(t[i]!=s[i])
+        {
+            extraChar[0] = t[i];
```
myrrlyn: Why a one-element array and not just the element?

challenge_8/C/randomDeepCopy.c
```diff
@@ -0,0 +1,111 @@
...
+#include <stdlib.h> 
+
+
+//node definition for the linked list
+struct node
+{
+	int data;
+	struct node* next;
+	struct node* random;
+};
```
myrrlyn: This is *completely irrelevant* to the rest of the review, just a comment.

Best practice is to order structs from widest to narrowest. Since pointers are always a machine word across, they are usually the widest and first (there are types wider than a word, but except on 32-bit CPUs those aren't common). I, personally, would order this next, random, data.

It's not really important, and the compiler will make it all work out, but ordering widest to narrowest will have the fewest paddings inserted. The compiler may or may not reorder to fill space if you have many partial-word items, which you don't here, but that's why it's a good practice.

challenge_8/C/randomDeepCopy.c
```diff
@@ -0,0 +1,111 @@
...
+
+typedef struct node* Node;
```
myrrlyn: You can also do this:

```c
typedef struct node {
  ...
} Node;
```

and do the layout and typedef in one shot.

challenge_8/C/randomDeepCopy.c
```diff
@@ -0,0 +1,111 @@
...
+
+//utility function to create a node
+Node newNode(int key)
+{
+	Node temp = (struct node*) malloc(sizeof(struct node));
```
myrrlyn: This should be `Node* temp`, not `Node temp`.

challenge_8/C/randomDeepCopy.c
```diff
@@ -0,0 +1,111 @@
...
+
+	temp->data = key;
+	temp->next = NULL;
+	temp->random = NULL;
+
+	return temp;
+}
+
+//utility function to print the list
+void printList(Node head)
```
myrrlyn: This should also be `Node*`, not `Node`, since you're accessing it as a pointer rather than a literal in the function.

challenge_8/C/randomDeepCopy.c
```diff
@@ -0,0 +1,111 @@
...
+	while(temp)
+	{
+		old = temp->next;
+		temp->next = old->next;
+		temp = old->next;
+		if(old->next) old->next->next;
+	}
+
+	return newNode;
+}
```
myrrlyn: You also mix immediate and referenced objects in here. `Node tmp` is accessed with `tmp.field`; `Node* other` is accessed with `other->field`.

`Node temp` is absolutely correct, but you may want to double check what should be references (`Node*`) vs immediates (`Node`) in this.

challenge_5/C/karanchawla/src/challenge5.c
```diff
@@ -0,0 +1,109 @@
...
+*/
+
+#include <stdio.h>
+#include <stdlib.h> 
+#include <time.h> 
+#include <string.h>
+
+
+//utility function used by qsort to sort the strings
+int comp (const void * elem1, const void * elem2)
```
karanchawla: http://www.cplusplus.com/reference/cstdlib/qsort/
`int compar (const void* p1, const void* p2);`
The function prototype used by qsort is like so^. Hence, I have to cast it from `void*` to `char*`. If I just define a method to compare two characters `qsort` doesn't work.

challenge_5/C/karanchawla/src/challenge5.c
```diff
@@ -0,0 +1,109 @@
...
+
+//function to shuffle the string
+void shuffle(char *array, size_t n)
+{
+    if (n > 1) 
+    {
+        size_t i;
+        for (i = 0; i < n - 1; i++) 
+        {
+            size_t j = i + rand() / (RAND_MAX / (n - i) + 1);
```
karanchawla: You're correct `i < n` instead of `n-1` and I can express `j` as `size_t j = rand()%n ;` too. It's just to get a random index <= size of the string.

challenge_5/C/karanchawla/src/challenge5.c
```diff
@@ -0,0 +1,109 @@
...
+    int size = sizeof(s)/sizeof(*s);
+    char t[size + 1];
+    strcpy(t,s);
+    
+    //edge case
+    if(size==0)
+        continue;
+    else
+        shuffle(s,size-1);
+
```
karanchawla: Yeah my code threw an error - maybe I wrote continue as an afterthought and never really ran the program after that. Thanks I made this correction.

challenge_5/C/karanchawla/src/challenge5.c
karanchawla: Hah yeah I added 'e' earlier but changed the test case later. Changed it.

challenge_5/C/karanchawla/src/challenge5.c
karanchawla: Yeah I don't know why? That's stupid - I just changed it to a `char`.

challenge_8/C/randomDeepCopy.c
karanchawla: Yes that's right but if I have renamed `struct node* node` as `Node` then I should use `other->field` right if `other` is of type `Node`. 
I have made the change though as you suggested i.e. 
`typedef struct node 
{

}Node;` and changed my `Node`'s to `Node*`

challenge_8/C/randomDeepCopy.c
karanchawla: I didn't get this, can you explain this a little further?

challenge_8/C/randomDeepCopy.c
karanchawla: PS I would like to thank you for taking the time out to review my code so meticulously. :smile: 
Thank you. Hopefully to more reviews this year 
Have a good one!

karanchawla:
Made the changes @myrrlyn 
Let me know if anything needs to be changed.

challenge_5/C/karanchawla/src/challenge5.c
myrrlyn: Ohhhhhh, that makes sense. Urgh, using `void` as a generic type hurts my brain.

myrrlyn:
Also, I renamed your files to fit our naming scheme; they are now in `challenge_{5,8}/c/karanchawla`. Lower-cased language, and in ch8 your work was top-level instead of in a personal folder. Pull, squash that warning, and you should be all set! Looks good on my end.

karanchawla:
Hey thanks for renaming the files, will keep that in mind for future challenges. Also why does it not give me that warning? 
image

-Karan

PR #316, opened 2017-01-09
Go challenge 2

makernaren:


challenge_2/go/makernaren/single_number.go
```diff
@@ -0,0 +1,34 @@
...
+		occurances[key]++
+	}
+	// Return when we find the single occurance.
+	for k, v := range occurances {
+		if v == 1 {
+			result = k
+			break
+		}
+	}
+	return result
```
erocs: I would recommend adding an additional 'error' return value that can be provided if no result was found.

makernaren: Thanks will change in next commit :)

PR #317, opened 2017-01-09
[Julia] Challenge_5 (Unreviewed)

dtcitron:
Julia language solution.

Also includes solution written in python/DTCitron


challenge_5/python/DTCitron/FindTheDifference.py
```diff
@@ -0,0 +1,23 @@
+class Solution:
+    def findTheDifference(self, s,t):
+        # Find the difference between the two lower-case strings s and t, where t has one additional randomly-inserted character
+        # First sort both s and t as lists of characters:
+        slist = sorted(list(s))
```
moonpatroller: Might not need to convert to list first here.

challenge_5/python/DTCitron/FindTheDifference.py
```diff
@@ -0,0 +1,23 @@
...
+        tlist = sorted(list(t))
+        # Now, just to be sure, let's check to see if s and t are equal:
+        if ''.join([i for i in slist]) == ''.join([i for i in tlist]):
+            return None
+        # Loop through all the items in tlist, comparing them one by one to the items in slist
+        for si in slist:
```
moonpatroller: Might be faster to just use indexes to compare array elements without popping.

PR #318, opened 2017-01-09
Python Challenge_4 (Unreviewed)

slansford:
Edited code to work correctly.


challenge_4/python/slansford/src/challenge_4.py
```diff
@@ -0,0 +1,66 @@
...
+        else:
+            if parent.right == None:
+                parent.right = node
+            else:
+                self._add(node, value, parent.right)
+
+
+#Inverts nodes of bTree-class object
+def invert(node):
+    toLeft = node.right
```
zmiller91: null pointer if node is `None`. This is valid, you should handle it: 

```
tree = bTree();
invert(tree.getRoot())
```

PR #319, opened 2017-01-09
ndsmith3's challenge_8 and challenge_9

ndsmith3:


challenge_8/scala/ndsmith3/WeirdLinkedList.scala
```diff
@@ -0,0 +1,87 @@
...
+    if (this.data == data) index
+    else this.next.get.indexOf(data, index + 1)
+  }
+
+  def subList(index: Int): WeirdList[A] =
+    (1 to index).foldLeft(this)((x, _) => x.next.get)
+
+  def foreach(f: WeirdList[A] => Unit): Unit = {
+    f(this)
+    if (this.next.isDefined) this.next.get.foreach(f)
```
moonpatroller: > if (this.next.isDefined) this.next.get.foreach(f)

Ouch!  Instead of checking if it's defined and then calling get, you should just use this.next.foreach(_.foreach(f)) which does nothing if this.next is None.  It's a little unfortunate that the 2 foreach's have the same name, but oh well.

challenge_8/scala/ndsmith3/WeirdLinkedList.scala
```diff
@@ -0,0 +1,87 @@
...
+    if (this.next.isDefined) this.next.get.add(data)
+    else this.next = Option(WeirdList(data))
+    if (this.length > 1) this foreach (x => assignRand(x)())
+    length += 1
+  }
+
+  def addNode(node: WeirdList[A]): Unit = {
+    if (this.next.isDefined) this.next.get.addNode(node)
+    else this.next = Option(node)
+  }
```
moonpatroller: > if (this.next.isDefined) this.next.get.addNode(node)
> else this.next = Option(node)

isDefined and get are code smells.  I would see if this works:

this.next.foreach(_.addNode(node)).orElse(this.next = Option(node))

ndsmith3:
@moonpatroller Fixed. Thank you for those suggestions. I am very new to using monads like this. :x

ndsmith3:
Just read about the new branch submission policy. Let me know if I need to reformat for these challenges.

ManuelMeraz:
@moonpatroller Do you think this is ready to merge?

challenge_10/scala/ndsmith3/ValidClosers.scala
```diff
@@ -0,0 +1,23 @@
...
+      lazy val continueWithPop  = innerClosers(in.tail, stack.tail)
+
+      in match {
+        case '(' :: _ | '[' :: _ | '{' :: _                => continueWithPush
+        case ')' :: _ | ']' :: _ | '}' :: _ if stackCloses => continueWithPop
+        case Nil                                           => stack.isEmpty
+        case _                                             => innerClosers(in.tail, stack)
+      }
+    }
+    innerClosers(str.toCharArray.toList)
```
moonpatroller: You should be able to use just str.toList because of the implicit? classes StringOps and StringLike.

challenge_8/scala/ndsmith3/WeirdLinkedList.scala
```diff
@@ -0,0 +1,103 @@
...
+    this.next match {
+      case None           => this.next = Option(node)
+      case Some(nextNode) => nextNode.addNode(node)
+    }
+  }
+
+  def indexOf(data: A, index: Int = 0): Int = {
+    if (this.data == data) index
+    else this.next match {
+      case None           => throw new Exception("Element not found")
```
moonpatroller: You should probably return an Option instead of throwing an Exception.

challenge_9/scala/ndsmith3/Squares.scala
```diff
@@ -0,0 +1,38 @@
+object Squares {
+  // Lazy values are used in the sortSquares function to avoid exceptions from
```
moonpatroller: This solution is pretty convoluted.  Yo could just use 2 indexes to work your way from the outsides in on the source array, copying to the target array.

moonpatroller: @ManuelMeraz I say merge it if @ndsmith3 agrees.

PR #321, opened 2017-01-09
[Python] Challenge 2

tavery321:


challenge_2/python/tavery321/src/c2.py
```diff
@@ -0,0 +1,8 @@
+arr = ['a', 'a', 'b','b', 1, 2, 2, 3, 3, 'z', 'z']
+answer = ''
+for i in range(len(arr)):
+	if(arr.count(arr[i]) == 1):
```
zmiller91: Your solution should run in O(n) time; this runs in O(n^2). Reconsider your usage of `arr.count(arr[i])` python has to look through your entire array to do this. Is there any way you can count all the elements with a single iterate of `arr`?

tavery321: Would the fix I made make it O(n)?

challenge_2/python/tavery321/src/c2fix.py
```diff
@@ -0,0 +1,12 @@
+a = ['a', 'b', 'c', 2, 3, 5, 'a', 2, 'c', 5, 'b']
```
moonpatroller: For lists you only use as a set, where you only add, remove, test for membership, and don't need the elements to stay in a certain order, it would be better to use a set or dict instead.

PR #326, opened 2017-01-10
[Python2]challenge_[9](Unreviewed)

shaowenliu:


challenge_9/python/shaowen/square_sort.py
```diff
@@ -0,0 +1,22 @@
...
+"""
+Created on Tue Jan 10 08:25:26 2017
+
+@author: Shaowen Liu
+"""
+
+def square_sort(arr):
+    # retieve the negtive and non-negtive list,O(n)
+    a_list = [-i for i in arr if i < 0]
+    b_list = arr[len(a_list):][::-1]
```
erocs: You don't need to use the second slice.  `b_list = arr[:len(a_list)-1:-1]`

challenge_9/python/shaowen/unit_test.py
```diff
@@ -0,0 +1,64 @@
+# -*- coding: utf-8 -*-
```
erocs: Add a she-bang line to start the file?
```
#!/bin/env python
```

challenge_9/python/shaowen/square_sort.py
shaowenliu: Nice one! Tks!

challenge_9/python/shaowen/unit_test.py
shaowenliu: Tks for the suggestion!
I just added, 
I never noticed this before, but in my understanding, this is only good for my own use, right? 
Cus the path is specific for my rig. If others want to run in './blabla.py' way, they have to change to their own path.

challenge_9/python/shaowen/square_sort.py
shaowenliu: just let you know, 
arr[:len(a_list)-1:-1] is sure efficient, but in my origin code, len(a_list) could be zero, in this case it will be b_list = arr[:-1:-1],  =>> b_list = [ ]. 
This's why, in the new commits, I treat the case (len(a_list) == 0) separately.

shaowenliu: just let you know,
arr[:len(a_list)-1:-1] is sure efficient, but in my origin code, len(a_list) could be zero, in this case it will be b_list = arr[: -1 : -1], =>> b_list = [ ].
This's why, in the new commits, I treat the case (len(a_list) == 0) separately.

challenge_9/python/shaowen/unit_test.py
shaowenliu: sorry, I just noticed the yearofprogramming home page is updated, it seems there are some new rules out there(especially the testing part)...i will read through it and make new commits later.

moonpatroller: > I never noticed this before, but in my understanding, this is only good for my own use, right?

No, there are fairly standard paths for these things.

PR #327, opened 2017-01-10
Solve Challenge #3 in Rust

myrrlyn:
This uses the same general setup as my Challenge 2 solution to find the majority element.

Requires a firm majority (> 0.5) to succeed, otherwise it fails.

By fails I mean "reports no majority", not crashes.


myrrlyn:
Man it feels silly to review my own submissions.

PR #333, opened 2017-01-10
[python]challenge_2 (Unreviewed)

sarcodian:


challenge_2/python/sarcodian/src/challenge_2.py
```diff
@@ -0,0 +1,12 @@
...
+    duplicate = []
+    for i in array0:
+        if i in duplicate:
+            pass
+        elif i in unique:
+            duplicate.append(i)
+            unique.remove(i)
+        else:
+            unique.append(i)
+    return unique[0]
```
moonpatroller: Using a set or dict for duplicates would be better because you expect most elements to go in there and scanning through a list to check if an element is in there and to remove an element is more expensive than doing those with a dict or set.

sarcodian:
Thanks moonpatroller, makes sense, haven't written anything longer then 20 lines in my life so don't have the habit of thinking about cost.

PR #334, opened 2017-01-10
[Python 3.6] Solution to challenge 10

Atakyn:


challenge_10/python/Atakyn/c10.py
```diff
@@ -0,0 +1,30 @@
...
+
+        if parenthesis < 0 or braces < 0 or brackets < 0:  # someone closed before opening
+            return False
+
+    if parenthesis == 0 and braces == 0 and brackets == 0:  # zero means as many openings as closings
+        return True
+    else:
+        return False
+
+print(valid_closers(input('Input the strings to test: ')))
```
zmiller91: Your solution fails with this test case: `{(})`

Atakyn: > Any characters can appear between any of these

Both open and close, what's wrong with that case?

moonpatroller: @zmiller you guys should fix the readme if you want that case to fail.  It actually reads like {(}) might be ok.

zmiller91: I did not write the problem, nor do I have the ability to change it. The problem is basic, a logical sequence of openers and closers must exist. `{(})` will not compile in any language.

PR #335, opened 2017-01-10
Elixir Challenge 2, 3, and 5 (Unreviewed)

joegotflow83:


challenge_3/elixir/joegotflow83/majority_element/lib/majority_element.ex
```diff
@@ -0,0 +1,26 @@
...
+    |> Map.to_list
+    |> determine_majority(length(input))
+  end
+
+  def determine_majority([head | tail], n) do
+    {num, value} = head
+    cond do
+      value > (n / 2) ->
+        num
+      true ->
```
myrrlyn: Why `true` and not `_`? Honest question; I'm just starting to learn Elixir so I'm not quite fluent yet.

moonpatroller: One of the cond cases has to evaluate to true.  It's the equivalent of an else block.

joegotflow83: @myrrlyn What moonpatroller said haha, the _ is used in case statements and stands for if none of the other statments can be pattern matched to the case then execute this code. Since cond is evaluating whether a statment is true or not, it needs one statement that is always going to be true.

myrrlyn: Good to know. I'm still wrapping my head around `case` vs `cond`. I would have expected `_` to be a truthy value, akin to a final `else` clause in an if stack or a `default` in a switch stack.

PR #340, opened 2017-01-10
[Python] Challenge 9 (Unreviewed)

wOstensen:


challenge_9/python/wost/src/main.py
```diff
@@ -0,0 +1,29 @@
...
+
+    result = []
+
+    while pos != [] and neg != []:
+        if pos[0] > neg[0]:
+            result.append(neg.pop(0))
+        else:
+            result.append(pos.pop(0))
+
+    return result + pos if neg == [] else neg
```
zmiller91: Your solution fails for test cases `[-2, -3]` and `[-3, 0, 1]`.

wOstensen:
It should work now! `[-2, -3]` is not a sorted list and therefore nothing my program needs to concern itself with.

challenge_9/python/wost/src/main.py
wOstensen: It should work now!
`[-2, -3]` is not a valid list, as it is not sorted.

PR #341, opened 2017-01-10
[x86] Challenge 10 (Unreviewed)

x95:
I hope I got the new workflow right


challenge_10/x86/x95/README.md
```diff
@@ -0,0 +1,16 @@
...
+##Compiling
+```
+$ as brackets.s -o brackets.o
+$ gcc -m64 challenge10.c brackets.o -o c10
+```
+tested on Ubuntu 16.04 64-Bit
+##Testing
+To test the implementation just start the c10 application and type the testcases in.
+
+The C file will call the asm function. Please take note, that this ASM file follows the AMD64 calling convention and not the Microsoft calling convention. Therefore it can only work on Linux/OSX/*NIX.
```
myrrlyn: MinGW toolchain on Windows uses the GNU ABI, and should work there. But yeah MSVC won't like it at all.

challenge_10/x86/x95/src/challenge10.c
```diff
@@ -0,0 +1,22 @@
...
+	char* inputbuf;
+	int check = 0;
+	while(1)
+	{
+		inputbuf = calloc(1024,sizeof(char));
+		printf("Input: ");
+		fgets(inputbuf,1024,stdin);
+		check = check_brackets(inputbuf);
+		printf("%d\n",check);
+		free(inputbuf);
```
myrrlyn: THANK YOU for `free()`ing your memory

PR #346, opened 2017-01-10
[Julia] Challenge 6 (Unreviewed)

dtcitron:
Julia language solution to Challenge 6

Also includes python solution, with unit test

### DIFF
--- a/challenge_6/python/DTCitron/src/main.py
+++ b/challenge_6/python/DTCitron/src/main.py
@@ -1,0 +1,40 @@
+def rangefinder(t):
+    # List of ranges
+    ranges = []
+
+    # Start at the very beginning
+    rangestart = t[0]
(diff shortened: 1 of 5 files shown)


challenge_6/python/DTCitron/src/main.py
```diff
@@ -0,0 +1,48 @@
...
+    if rangeend > rangestart:
+        ranges.append(str(rangestart) + '->' + str(rangeend))
+    return ranges
+
+
+if __name__=="__main__":
+    t1 = [1,2,3,4,8,9,10,12,13,14]
+    t2 = [1,2,3,4,9,10,15]
+    t3 = [1,2]
+    t4 = [1]
```
moonpatroller: Should probably check [1, 4] too.

PR #350, opened 2017-01-11
Challenge_3 in Python (Unreviewed)

ajschrier:


challenge_3/python/ajschrier/README.md
```diff
@@ -0,0 +1,5 @@
+# Challenge 3 - Majority Element
+
+This uses the list.count builtin to find out how many of each element exists in the list. If the given element meets the condition (count > len/2), then that element is printed to the console and the loop breaks. 
+
+I feel like I save some time by breaking the loop but there may be a more efficient way to do this. I'll probably revisit this in the future.
```
zmiller91: There's definitely a better way to do it; it's not because you iterate over the entire array.  Consider what `array.count(i)` does and how many times you are finding the counts. Is there a way you can count everything upfront, with one pass of `array`?

PR #351, opened 2017-01-11
[Python2] Challenge_[10](Pending)

shaowenliu:


challenge_10/python/shaowen/solution.py
```diff
@@ -0,0 +1,27 @@
...
+
+@author: Shaowen Liu
+"""
+def solution(strr):
+    """check if all brackets closed properly"""
+    
+    closerof = {'{':'}', '[':']','(':')'}
+    stack = []
+    
+    for i in list(strr):
```
erocs: No need to convert the string to a list. Iterating over a string provides the characters.

challenge_10/python/shaowen/solution.py
```diff
@@ -0,0 +1,27 @@
...
+        # push into stack, if new bracket found
+        if i in ['[','{','(']:
```
erocs: You already have a map of the starter characters. How about `if i in closerof:`?

challenge_10/python/shaowen/solution.py
```diff
@@ -0,0 +1,27 @@
...
+            stack.append(i)
+        # pop out if proper closer found, else False
+        elif i in [']','}',')']:
```
erocs: For constant arrays like `[']','}',')']`, prefer using a tuple.  `(']','}',')')`  Also, make it a module level constant so it doesn't need to be initialized every function call.

challenge_10/python/shaowen/solution.py
```diff
@@ -0,0 +1,27 @@
...
+            if closerof[stack[-1]] == i:
+                stack.pop()
+            elif closerof[stack[-1]] !=i:
```
erocs: This elif condition is unnecessary as an else covers the opposite of your if expression without needing to perform another test.

challenge_10/python/shaowen/solution.py
```diff
@@ -0,0 +1,27 @@
...
+                return False
+    # check leftover in stack
+    if stack:
```
erocs: Instead of this if / else, you can just `return not stack`.

challenge_10/python/shaowen/unit_test.py
```diff
@@ -0,0 +1,37 @@
...
+@author: Shaowen Liu
+"""
+import unittest
+from solution import solution
+
+class solution_test(unittest.TestCase):
+    
+    def test_case1(self):
+        strr = '{{{}}}'
+        self.assertEqual(solution(strr), True)
```
erocs: Instead of using `assertEqual` to test True / False conditions, use `assertTrue` and `assertFalse`.

challenge_10/python/shaowen/unit_test.py
```diff
@@ -0,0 +1,37 @@
...
+        self.assertEqual(solution(strr), False)
+        
+    def test_case5(self):
+        strr = '{{{{{{{{{}}}}}}}}}'
+        self.assertEqual(solution(strr), True)
+
+    def test_case6(self):
+        strr = '[[[[[[[[[kafjalfeianfailfeja;fjai;efa;sfj]]]]]]]]]kjajdain'
+        self.assertEqual(solution(strr), True)
+
```
erocs: Add a test case for `'([{)}]'` to verify out of order closers aren't accepted.

challenge_10/python/shaowen/solution.py
shaowenliu: tks!
In my mind, string is not an efficient date structure, so I used to convert it to list......
Well, in this case, absolutely not necessary!

challenge_10/python/shaowen/solution.py
shaowenliu: Exactly!!!

challenge_10/python/shaowen/solution.py
shaowenliu: clever shot! thanks man!

challenge_10/python/shaowen/unit_test.py
shaowenliu: Yep!

challenge_10/python/shaowen/unit_test.py
shaowenliu: Got it!
Never try that function before!

challenge_10/python/shaowen/solution.py
shaowenliu: I think I will make it tuple constants for both ( '[' , '{' , '(' ) and ( ']' , '}' , ')' )  , as you in your next suggestion. it should be easier to understand.

shaowenliu:
@erocs, 
Appreciate the advice! 
I did the modification and now the code looks clearcut.

PR #352, opened 2017-01-11
[Python] Challenge 9 (ReadyForMerge)

alexbotello:
This one took me awhile and probably would of taken longer if not for the discussion on slack :)

### DIFF
--- a/challenge_9/python/alexbotello/src/square.py
+++ b/challenge_9/python/alexbotello/src/square.py
@@ -2,7 +2,6 @@ def square_soft(input_list):
     """
     Returns a list with all values squared and sorted
     """
-
     squared = []
 
     # Separate negative and positive values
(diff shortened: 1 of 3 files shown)


challenge_9/python/alexbotello/src/square.py
```diff
@@ -0,0 +1,40 @@
...
+    pos = [x**2 for x in input_list if x >= 0]
+    neg.reverse()
+
+    # While loop will exit when squared takes in every value from input_list
+    # pop() each value after appending as to avoid using a counter variable
+    # and possibily getting an IndexError
+    while len(squared) != len(input_list):
+        # Add the rest of pos array if neg is empty
+        if not neg:
+            squared.append(pos.pop(0))
```
halogenandtoast: instead of just appending could you do:

```python
squared.extend(pos)
```

You would then incur less iterations of your while loop.

challenge_9/python/alexbotello/src/square.py
```diff
@@ -0,0 +1,40 @@
...
+        # Add the rest of neg array if pos is empty
+        elif not pos:
+            squared.append(neg.pop(0))
```
halogenandtoast: Same here.

challenge_9/python/alexbotello/src/square.py
```diff
@@ -0,0 +1,40 @@
...
+
+        elif pos[0] >= neg[0]:
+            squared.append(neg.pop(0))
+
+        elif neg[0] >= pos[0]:
```
halogenandtoast: Is this conditional necessary. At this point I think you could just use else

PR #356, opened 2017-01-11

sarcodian:


whiterd:
@slandau3 Could you elaborate on what you mean by "Make it so the main function in the script can be activated straight away. Without us having to call it."?  I'm really new to a lot of terminology.

whiterd:
@slandau3 Thank you. That makes perfect sense now. I have a greater appreciation for docstrings after bouncing around different scripts.

sarcodian:
Hello, added the documentation and made it autorun.

jarydremillard:
Bypass

PR #358, opened 2017-01-11
[Python] Challenge 4 (Unreviewed)

sarcodian:


ManuelMeraz:
@sarcodian Can you scale this up to any size binary tree? I think I would be able to accept this PR if I could see an tree inversion function for a binary tree of any size, not just the given example. The way you have your array should be fine, just make it so it scales up.

sarcodian:
Hey, added a method to make random trees.

ManuelMeraz:
@sarcodian Would your code work if we had a binary tree with 128 integers with a depth of lg128 = 7? What about with 256?

sarcodian:
@ManuelMeraz It should. the actual reverse implementation is the same logic as some/most of the other submissions, recurse down until you get a null. Although not from the CS field so I have no idea if you would run into resource issues with this. I am just stubbornly sticking to lists instead of making a class in data/left/right format.

jarydremillard:
@sarcodian In a large scale, you would lose memory really really quick. But for now, I see this a slightly unique, albeit not the best, solution. With that said, Ill accept it.

PR #361, opened 2017-01-11
[Python] Challenge 7 (Unreviewed)

sarcodian:


challenge_7/python/sarcodian/src/challenge_7.py
```diff
@@ -0,0 +1,12 @@
+def missing_int(L):
+
+    sum_n = 0
+    sum_n_less_1 = 0
+
+    for i in range(len(L)+1):
+        sum_n += i
+    
+    for i in L:
+        sum_n_less_1 += i
```
ningyuansg: Try the more pythonic and readable `sum(L)`.

challenge_7/python/sarcodian/src/challenge_7.py
```diff
@@ -0,0 +1,12 @@
+def missing_int(L):
+
+    sum_n = 0
+    sum_n_less_1 = 0
+
+    for i in range(len(L)+1):
+        sum_n += i
```
ningyuansg: Try the more pythonic `sum([i for i in range(len(L)+1)])`

sarcodian: Thanks, can you provide a link or something that provides a deeper explanation of <i for i in range(len(L)+1)> syntax. I have seen it a lot in other code as well, but I don't get the logic behind how it works so I am wary of just using it. Or even if you provide the name for what that type of syntax is called I can google it.

challenge_7/python/sarcodian/src/challenge_7.py
sarcodian: Haha, thanks, didn't know it existed :)

challenge_7/python/sarcodian/src/challenge_7.py
ningyuansg: As sent to you in slack, here's a great video on the topic (and also for the reference of others)
https://www.youtube.com/watch?v=ZoWgzG_r2qo

PR #363, opened 2017-01-11
[Python] Challenge 9 (Unreviewed)

sarcodian:


challenge_9/python/sarcodian/src/challenge_9.py
```diff
@@ -0,0 +1,23 @@
...
+
+    L_square = []
+    L_sorted = []
+    count = len(L)
+
+    if L[0] >= 0:
+        for i in L:
+            L_square.append(i**2)
+        return L_square
+
```
shaowenliu: is the first 'if' step necessary? it seems to me,, the later 'while' loop is enough to kill the problem.

sarcodian: I know it isn't, but trying to practice / try to be efficient. I figure the while loop is only needed with negative numbers. so if the numbers are all positive, then why bother comparing and appending, and then reversing. Imagine if the list had 500 positive numbers.

PR #364, opened 2017-01-11
[Python] Challenge 10 (Unreviewed)

sarcodian:


karanchawla:
You haven't considered '<' and '>' in your method. Also add documentation for your function definition.

sarcodian:
The < and > weren't in the original spec and as such are ignored like other char's.

PR #369, opened 2017-01-11
[Python] Challenge 5 submission

whiterd:


whiterd:
I'm not sure how challenge 4 and 5 got into the same pull request.  Should I delete this and resubmit it?

ningyuansg:
@whiterd Sure. Or try:

```
$ git checkout python_5
$ git rm -r challenge_4/python/whiterd
$ git commit -m "Remove challenge #4"
```

ningyuansg:
Note: Your challenge 4 is in this pull request because it is in this, your challenge 5 branch. As you know, this shouldn't be the case, as each challenge is supposed to have its own branch. You could've produced this error in two ways (that I know off)

1. When you started off in your branch `origin/master`, you already had challenge 4 in there. Therefore when you created this new branch `python_5` it was included as well. However, as your challenge had not yet been merged with with this `challenges/master` branch, it shows up as a *_change_* in the pull request.
2. You created the branch correctly with a clean `origin/master` without a challenge 4. However, you somehow merged your `python 4` with `origin/master`. Afterwards, while on your branch `python_5` you performed a `git pull origin master`, merging challenge 4 content into this branch.

To avoid this, simply make sure the `origin/master` branch is not ahead of `challenges/master`. Do this by only calling `git pull challenges master`; avoid pulling from your `origin/python_n`. You must also not pull anything into your `origin/python_n`.

PR #370, opened 2017-01-11
[Julia] Challenge 7 (Unreviewed)

dtcitron:
Also includes a solution in python


challenge_7/python/DTCitron/src/main.py
```diff
@@ -0,0 +1,16 @@
+import numpy as np
+
+def findMissing(l):
+    n = len(l)
+    return n*(n+1)/2 - np.sum(np.array(l))
```
moonpatroller: You don't need numpy to use sum.

PR #375, opened 2017-01-11
[C++] Challenge 10 (Unreviewed)

ManuelMeraz:


challenge_10/cpp/manuel/src/tools.cpp
```diff
@@ -0,0 +1,45 @@
...
+    int length = input.length();
+
+    for(int i = 0; i < length; i++) {
+        if(input[i] == '(' || input[i] == '{' || input[i] == '[' || input[i] == '<') {
+            // count all the openers
+            counter[input[i]]++;
+
+            // count the closers
+        } else if(input[i] == ')') {
+            counter['(']--;
```
moonpatroller: Would this work correctly for "{ [ } ]" ?

PR #376, opened 2017-01-11
[Python] Challenge 6 submission

whiterd:
I feel rather good about figuring this one out, but the last append outside of my  for loop just doesn't feel right.  I would appreciate any feedback.


Loran425:
Looks good for the most part. There is an edge case where a single number is not a range. so 11->11 would be an invalid output. but other than that your code is looking good. 

I had to implement the end of loop thing also. Another way you could do it is append a 0 to end of the list of numbers so you know it will fail and run your append condition at the end of a list. 

To make life easier as you go further you might look into unit tests. These make it easy to run identical tests when you change your code accounting for edge cases and the like. 

- [ ] **Code Correctness**
- [X] **Code Style**
- [X] **Code Documentation**
- [X] **Path Structure**
- [ ] **Merge Pull Request**

PR #378, opened 2017-01-11
[C++] Challenge 9 (Unreviewed)

ManuelMeraz:


challenge_9/cpp/manuel/src/tools.cpp
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+
+    while(right_length > 0) {
+        numbers[right_index] = pow(numbers[right_index], 2);
+        sorted.push_back(numbers[right_index]);
+        right_index++;
+        right_length--;
+    }
+
```
karanchawla: A comment here too.

ManuelMeraz: Done!

PR #379, opened 2017-01-11
[Java] Challenge 10

jdfurlan:
Still figuring out the right way to branch etc. any feedback is welcome


challenge_10/java/jdfurlan/src/ValidClosers.java
```diff
@@ -0,0 +1,68 @@
...
+			return true;
+		List<Character> openers = Arrays.asList('{', '[', '(');
+		List<Character> closers = Arrays.asList('}', ']', ')');
+		Stack<Character> stack = new Stack<Character>();
+		for (char c : nextLine.toCharArray()) {
+			// if we found an opener just put it at the top of the stack
+			if (openers.contains(c)) {
+				stack.push(c);
+			}
+			if (closers.contains(c)) {
```
moonpatroller: else if

challenge_5/java/jdfurlan/src/FindTheDifference.java
```diff
@@ -0,0 +1,21 @@
+
+public class FindTheDifference {
+
+	public static void main(String[] args) {
+		String s = "abcd";
+		String t = "dbeca";
+
+		boolean[] b = new boolean[128];
+		for (int i = 0; i < s.length(); i++) {
+			b[s.charAt(i)] = true;
```
moonpatroller: Does this handle the same character being in an input string twice or more?

jdfurlan: Yes it does. eg. `String s = "apple"; String t = "plepva";` output is `v found at index: 4` the reason being every character, if found, is stored in its char value index in a boolean array. regardless of how many are present in the string, it will will see `true` for that char and know that it was present in the original string.

challenge_10/java/jdfurlan/src/ValidClosers.java
jdfurlan: good catch thanks!

PR #386, opened 2017-01-12
challeng7 in python

py2k5:


challenge_7/python/py2k5/findMissingNumber.py
```diff
@@ -0,0 +1,14 @@
+
+def main():
+	numbers = [ 0,1,3,4,2,5,7 ]
+        tot = len(numbers)
+	if 0 not in numbers: 
+		print 0
+	else:
+		for num in numbers:
+			if num +1 < tot and num+1 not in numbers:
```
shaowenliu: 'num + 1 not in numbers is O(n),  and you put it in the for loop, together it's O(n square) complexity?

ManuelMeraz:
@slansford 
- [ ] **Code Correctness**
Your challenge 7 solution needs to be worked on. I think it's just a misunderstanding of how the statement `if i in list` works. To confirm whether something is in a list this statement will iterate through your list until it finds the number, then check the next number and iterate through the list again, until the first for loop finishes. In a sense you have 2 nested for loops each dependent on the length of the input to find the missing number making this O(N^2).  See if you can improve on this algorithm. 
- [ ] **Code Style**
- [x] **Code Documentation**
- [x] **Path Structure**
- [ ] **Merge Pull Request**

py2k5:
I am still learning python. :)

ManuelMeraz:
@py2k5 We can help you in the slack chat if you'd like or I can give you some hints here as well on how to improve. I'd also recommend looking at some of the other python solutions for this challenge.

py2k5:
sure in fact I wll be glad. let me know when we can get into a call.

PR #390, opened 2017-01-13
[Python] Challenge 7 (Pending)

slansford:


ManuelMeraz:
- [ ] **Code Correctness**
- [ ] **Code Style**
- [ ] **Code Documentation**
- [ ] **Path Structure**
- [ ] **Merge Pull Request**

challenge_7/python/slansford/src/challenge_7.py
```diff
@@ -0,0 +1,13 @@
+def missingNo(inputList):
+
+	minInt = min(inputList)
+
+	if minInt != 0:
+		return 0
+
+	for i in inputList:
+		if minInt in inputList:
```
ningyuansg: > O(N) times means that your algorithm can not contain nested loops

The average for `x in list` look up in python is O(n). This makes this solution O(n^2).

source: https://wiki.python.org/moin/TimeComplexity

PR #395, opened 2017-01-14
[Python]challenge_12(Unreviewed)

shaowenliu:


shaowenliu:
It seems that, this request includes the previous pull request,, maybe someone has to delete the previous one?(pr #385 is still there, overlapped with this new one)
I'm trying/leaning to push a separate request for the challenge...but failed this time.

shaowenliu:
It is encouraged to have a separate pull request for each challenge. right?
 Or, it's totally fine to pull together?

PR #397, opened 2017-01-14
[Python] Challenge #10 (Pending)

ningyuansg:
I found three cases where the sequence can be considered to have 'invalid closers'.

1. A mismatch in the number of opening and closing brackets, e.g. `())`
2. Premature closing, e.g. `][`
3. Overlapping of brackets, e.g. `([<)]>`

In this solution, we use a list to 'count' the brackets as we iterate through the list. Every opening bracket adds to this list `mem`, and closing brackets remove their corresponding openings from the list.

The first invalid case is partially tested for by checking if the list `mem` is empty at the end of the iteration. The works for an abundance, but not shortage of opening brackets.

For the case whereby there is a shortage of closing brackets, we simply return False in the case of an `IndexError`. This also covers the second case of premature closing.

For the third case of overlapping, we check for every closing bracket if the most recent _unclosed_ opening bracket is of the same type. If not, False is returned as expected.

The unit test, `test.py`, is constructed simply from the examples given.

```
.........
Ran 9 tests in 0.001s

OK
```

### DIFF
--- a/challenge_10/python/ning/README.md
+++ b/challenge_10/python/ning/README.md
@@ -1,0 +1,25 @@
+# Challenge #10, Valid Closers
+
+I found three cases where the sequence can be considered to have 'invalid closers'.
+
+1. A mismatch in the number of opening and closing brackets, e.g. `())`
+2. Premature closing, e.g. `][`
+3. Overlapping of brackets, e.g. `([<)]>`
+
+In this solution, we use a list to 'count' the brackets as we iterate through the list. Every opening bracket adds to this list `mem`, and closing brackets remove their corresponding openings from the list.
+
+The first invalid case is partially tested for by checking if the list `mem` is empty at the end of the iteration. The works for an abundance, but not shortage of opening brackets.
+
+For the case whereby there is a shortage of closing brackets, we simply return False in the case of an `IndexError`. This also covers the second case of premature closing.
+
+For the third case of overlapping, we check for every closing bracket if the most recent _unclosed_ opening bracket is of the same type. If not, False is returned as expected.
+
+The unit test, `test.py`, is constructed simply from the examples given.
+
+```
+.........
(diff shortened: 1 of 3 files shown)


ManuelMeraz:
@ningyuansg I'll change this PR to pending, thank you for reviewing

ManuelMeraz:
- [ ] **Code Correctness**
Your works just great. I just have a suggestion, have you considered a [dictionary](https://docs.python.org/2/tutorial/datastructures.html#dictionaries) for looking up values? It will run faster. For example instead of 
```
 if character in OPENS
```
can have a dictionary look up as
```
if OPENS.get(character) == None: 
    do this
```
A dictionary will allow you to instantly look to see if an item exists rather than using `if i in list` because `i in list` will iterate through your list and check one by one. If you'd like to implement a dict I can wait, otherwise I think your code is acceptable. 
- [ ] **Code Style**
Clean, I like it.
- [ ] **Code Documentation**
Well documented.
- [ ] **Path Structure**
- [ ] **Merge Pull Request**

ManuelMeraz:
@ningyuansg

ningyuansg:
@ManuelMeraz 

Thanks for the review. I didn't know about the use of dictionaries as look up, I'd be happy to learn more about that. I assume you mean something like,

```
OPENS = {
    '(': '',
    '[': '',
    '<': '',
    '{': ''
}

if OPENS.get(character) != None:
    # code for opening brackets
else:
    # code for other characters
```

I'm using sets (and so, set look up) in my solution though. If I'm not wrong, that makes it much faster than list look up, and perhaps comparable to dictionary look ups in terms of speed?

ManuelMeraz:
@ningyuansg Ah okay, sounds great! I'll merge your solution. Set lookup is basically O(1).

PR #403, opened 2017-01-15
[Python] Challenge #12 (Pending)

ningyuansg:
Our solution comes in two parts, `compress` and `decompress`. Running `challenge_12.py` will invoke an 'interactive' mode that will compress or decompress according to the input.

---

`compress`,

1. The main logic in this function is to iterate through characters of the given input string, and compare it to the previous character to determine if it is a different character.
2. If it is a different character, we will examine how many times the previous character was repeated.
3. If there are greater than 4 repetitions, we append to the output string `compressed` in the given format {character}#{repetitions}. Otherwise, simply append with no changes.
4. To facilitate this, we store iterated strings in a buffer list `buff`. This list is reset after each application of the logic in (3).

Some minor features:

* The first character is added to the buffer before the iterating `for` loop to prevent `IndexError` for the comparison conditional detecting change in character `if character == buff[-1]`.
* There is a `else` for the main iterating `for` loop for the case whereby the string ends on a sequence of repeats (and therefore there is no differing character to trigger `elif character != buff[-1]`

---

`decompress`,

1. The main logic in this function is to iterate through the characters of the given input string, and append all alphabetical characters into the output string.
2. Numerical characters are held until they are followed by an alphabet (in the case of numbers greater than 9), after which an appropriate number of repetitions of the last alphabet is appended tot he output.

Some minor features:

* The '#' character is removed at the start as it serves no purpose (numbers are sufficient to indicate a sequence of compressed repetition).
* As with `compress`, an `else` block for the main iterating `for` loop catches the case whereby the input string ends in a number.

---

```
......
Ran 6 tests in 0.002s

OK
```

### DIFF
--- a/challenge_12/python/README.md
+++ b/challenge_12/python/README.md
@@ -2,6 +2,8 @@
 
 Our solution comes in two parts, `compress` and `decompress`. Running `challenge_12.py` will invoke an 'interactive' mode that will compress or decompress according to the input.
 
+---
+
 `compress`,
 
 1. The main logic in this function is to iterate through characters of the given input string, and compare it to the previous character to determine if it is a different character.
@@ -14,6 +16,8 @@ Some minor features:
 * The first character is added to the buffer before the iterating `for` loop to prevent `IndexError` for the comparison conditional detecting change in character `if character == buff[-1]`.
 * There is a `else` for the main iterating `for` loop for the case whereby the string ends on a sequence of repeats (and therefore there is no differing character to trigger `elif character != buff[-1]`
 
+---
+
 `decompress`,
 
 1. The main logic in this function is to iterate through the characters of the given input string, and append all alphabetical characters into the output string.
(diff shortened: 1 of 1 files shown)


ningyuansg:
_(note to self: learn regex...)_

ningyuansg:
[repl.it link](https://repl.it/FN6N/0) for an interactive mode so you can test this code out.

challenge_12/python/challenge_12.py
```diff
@@ -0,0 +1,72 @@
...
+            compressed += ''.join(buff)
+        buff = [character]
+    
+    return compressed
+
+def decompress(strr):
+    strr = strr.replace('#', '')
+    repeats = ''
+    decompressed = ''
+    NUMS = ('0', '1', '2', '3', '4', '5', '6', '7', '8', '9')
```
moonpatroller: Should this handle numbers over 9?

ningyuansg: It does!

Repl.it compiler running this code...

Multi-digit numbers are appended to the variable `repeats` until they end (i.e. a character is parsed). This conditional is at line 37, `if character not in NUMS:`, which will run line 43, `repeats += character`.

PR #405, opened 2017-01-15
[x86] Challenge 4 (ReadyForMerge)

x95:
Easy Code I promise


challenge_4/x86/x95/src/invertTree.s
```diff
@@ -0,0 +1,62 @@
...
+	# Traverse left node
+	cmp $0,%rsi
+	je skip_left
+	pushq %rdi	# Preserve %rdi and proceed with %rsi
+	movq %rsi,%rdi
+	call invertTree
+	popq %rdi
+	skip_left:
+
+	# Get addresses of both nodes
```
erocs: The swapping could be made simpler.
```
  movq 8(%rdi), %r8
  movq 16(%rdi), %r9
  movq %r8, 16(%rdi)
  movq %r9, 8(%rdi)
```

x95: @erocs I am always using Linux for my projects, but I will write it a bit more prominently next time. Thanks for the tip, I mainly wanted to play around with the exchange instruction when I coded it. Will remember it in the future

erocs: I know you use linux. It was I who didn't pay attention and sat here wondering why it wasn't working, lol.

No worries, always good to know other ways to do things.  :)

PR #408, opened 2017-01-16
[Cpp] Challenge 2 (Unreviewed)

rbrtdambrosio:


rbrtdambrosio:
@jacktraf Thanks for the comment :) I'm interested in this. Why is dangerous in C? I don't know C very well so I would like to understand more about it.

jacktraf:
Reading/writing past the end of a buffer/array in c/c++ is a big no no.  Its classified as undefined behavior / buffer overflow (in the case of writing). 
 "In C the use of any automatic variable before it has been initialized yields undefined behavior, as does integer division by zero or indexing an array outside of its defined bounds (see buffer overflow)" https://en.m.wikipedia.org/wiki/Undefined_behavior

See also   https://en.m.wikipedia.org/wiki/Buffer_overflow    for possible security risks

jacktraf:
Also, never trust user input

rbrtdambrosio:
@jacktraf thanks for the answer :)
Yes i'm aware of the issues regarding the use of array beyond their end or without initialization.
I was only assuming a "legal" behavior from the user. If an integer value (in the int range) is used as array length there shouldn't be problem.
In any case it's an interesting point that I will keep in  mind :)

PR #415, opened 2017-01-17

dfdx2:
Python Challenge 2 Unreviewed


jarydremillard:
@dfdx2 Although this technically works, and is a solution, this is geared more towards writing your algorithm instead of letting a library do it for you. Granted there are times when we use libraries to do so instead of reinventing the wheel, but I hope you understand my point.

Bypass

PR #420, opened 2017-01-17
[Python] Challenge 13 (Pending)

alexbotello:
We have our function take the integer input and convert it into a list of
integers. We find the number of digits, the midpoint, and then separate both
halves of the list.

If the number of digits is even, reverse the second half and compare it to the first. If number of digits are odd, follow the same steps, but be sure to add +1 to the index of the first half.


alexbotello:
Okay, I ditched the list implementation altogether and decided to reverse the int using %10

PR #424, opened 2017-01-17
[Python]Challenge_5(Reviewed)

ajschrier:


ajschrier:
@MJUIUC Thanks for the review and the suggestions. I use spaces around operators for PEP8 compliance

PR #425, opened 2017-01-17
[Python]Challenge_6(Reviewed)

ajschrier:


challenge_5/python/ajschrier/FindTheDifference.py
```diff
@@ -0,0 +1,21 @@
+from collections import Counter
+
+
+class Solution():
```
erocs: There's no need to use a class here as you are effectively defining a standalone method. Just make it a module level method.

challenge_6/python/ajschrier/IntRange.py
```diff
@@ -0,0 +1,41 @@
...
+
+def outputRange(inputList):
+    # Helper Vars
+    outputList = []
+    rangeStart = None
+    rangeEnd = None
+
+    for i in inputList:
+        # Detect the beginning of the list
+        if not rangeStart:
```
erocs: You will want to use `if rangeStart is None:`, as a zero value starting the list would be overridden with the second iteration.

challenge_5/python/ajschrier/FindTheDifference.py
ajschrier: Did that to match the unit test file that was provided, I agree about the style point though

challenge_6/python/ajschrier/IntRange.py
ajschrier: Nice catch, thanks

challenge_6/python/ajschrier/IntRange.py
```diff
@@ -0,0 +1,41 @@
...
+    rangeStart = None
+    rangeEnd = None
+
+    for i in inputList:
+        # Detect the beginning of the list
+        if rangeStart is None:
+            rangeStart = i
+            rangeEnd = i
+        # Check for invalid sequence
+        elif rangeEnd is not i-1:
```
erocs: `is` should be used for object identity testing not equality testing. Just use !=.

erocs: I should add, this differs from the above `if rangeStart is None` check because `None` is an explicit object that you are testing against it's direct identity. All None values are the exact same object.

PR #427, opened 2017-01-18
Go challenge 3

makernaren:


jarydremillard:
Why do you have two files;
max_occurance.go & max_occurance_test.go?

makernaren:
One is the source code, other is for testing it. I guess thats the standard format for writing tests for golang code. `yourCode_test.go`

challenge_3/go/makernaren/max_occurance.go
```diff
@@ -0,0 +1,38 @@
...
+// most number of times.
+package main
+
+import (
+	"fmt"
+)
+
+// Finds the maximum occurance, takes in interface as input and output params
+//as elements in the given array may be of any data type
+func FindMajority(input []interface{}) (result interface{}) {
```
erocs: A slice of empty interfaces is unwieldy. Sure, it's Go's iffy "generics", but if you have a typed slice, it has to be copied into this new generic slice before being passed in. You would likely end up just writing a FindMajorityTYPE for each type you wanted this functionality for or you'd create a new interface that better supported what this needs to work.

For this problem, it's ok, as you create the slice of empty interfaces as input and it's not used for anything else.

challenge_3/go/makernaren/max_occurance.go
```diff
@@ -0,0 +1,38 @@
+// Take an array of mixed data types and find the element that occurs
+// most number of times.
+package main
```
erocs: I would put the challenge functionality and unit tests in a challenge specific package and break out main() into its own file / package.

challenge_3/go/makernaren/max_occurance.go
```diff
@@ -0,0 +1,38 @@
...
+//as elements in the given array may be of any data type
+func FindMajority(input []interface{}) (result interface{}) {
+	// Create a hashmap, where element of that array is key and its number of
+	// occurances is the value.
+	occurances := make(map[interface{}]int)
+	for _, key := range input {
+		occurances[key]++
+	}
+	// Return when we find the max occurance.
+	max := 0
```
erocs: You can calculate this on the fly in the loop where you calculate occurrences. Just track what the current item / max count is and update that if the element you just incremented is larger.

challenge_3/go/makernaren/max_occurance.go
```diff
@@ -0,0 +1,38 @@
...
+	kv_pair := make([]interface{}, 2)
```
erocs: Using a pair to track is odd. Just create two separate local variables each with their own descriptive name.

PR #432, opened 2017-01-19
[Python]Challenge_7(Reviewed)

ajschrier:


challenge_7/python/ajschrier/README.md
```diff
@@ -0,0 +1,19 @@
...
+In a list 0..N, find which digit is missing.
+
+# Functions
+
+## builtInMethod
+
+**Input:** List
+**Output:** Integer
+
+Finds the integer 0 < x < N not present in list. The 'not in' syntax performs at O(N) according to [the Python Wiki](https://wiki.python.org/moin/TimeComplexity), so that is used to check the list for the value in the range.
```
erocs: By utilizing it in your xrange loop you are creating a O(n^2) solution. The first O(n) is your loop, which is multiplied by the nested O(n) of the `in` operator.

ajschrier: I knew I was missing something on that. Thanks

challenge_7/python/ajschrier/MissingNumber.py
```diff
@@ -0,0 +1,23 @@
+def builtInMethod(inputList):
+    for i in xrange(len(inputList)):
+        # Check if i is in list
+        if i not in inputList:
+            # Return number
+            return i
+    # Return 0 if none of the numbers meet the condition
+    return 0
```
erocs: It would be more Pythonic to raise an Exception. Also 0 is in the list of numbers so return -1.

challenge_7/python/ajschrier/MissingNumber.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def sumMethod(inputList):
+    # find (n*n+1)/2
+    fullSum = (len(inputList) * (len(inputList) + 1)) / 2
+    # subtract sum of full range from supplied range
+    return fullSum - sum(inputList)
+
+
+def main():
```
erocs: If main doesn't do anything you can delete from here down.

PR #439, opened 2017-01-20
[go] Challenge 1 (Unreviewed)

tml:


challenge_1/go/tml/src/strrev.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+    "fmt"
+    "os"
+    "bufio"
+)
+
+func strrev(s string) string {
+    r := []rune(s)
+    rs := []rune{}
```
erocs: Utilize make() to create the new rune slice with enough space to hold the string being reversed.

rs := make([]rune, 0, len(s))

challenge_1/go/tml/src/strrev.go
```diff
@@ -0,0 +1,23 @@
...
+    for i := len(r)-1; i>=0; i-- {
+        rs = append(rs, r[i])
```
erocs: Alternatively, iterating over the string's range provides each rune.

```
for _, ch := range s {
  rs = append(rs, ch)
```

challenge_1/go/tml/src/strrev.go
```diff
@@ -0,0 +1,23 @@
+package main
+
+import (
+    "fmt"
+    "os"
+    "bufio"
+)
+
+func strrev(s string) string {
```
erocs: You should capitalize methods you want exported from the package, as this makes it public. Starting an identifier with lower-case makes it package private.

PR #441, opened 2017-01-20
[Go] Challenge 1

shivam99aa:


challenge_1/go/shivams/reverse.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"fmt"
+	//"strings"
+)
+
+func reverseString(str string) string {
+	var revStr string
+	for i := len(str) - 1; i >= 0; i-- {
+		revStr = revStr + string(str[i])
```
erocs: This will cause a lot of memory churn as it copies revStr to a new instance for each character this appends to the end.

PR #444, opened 2017-01-20
[Java] Challenge 14 [ReadyForMerge]

jdfurlan:


challenge_1/java/jdfurlan/src/ReverseAString.java
```diff
@@ -0,0 +1,18 @@
+
+import java.util.Scanner;
+
+public class ReverseAString {
+
+	public static void main(String[] args) {
+		Scanner sc = new Scanner(System.in);
```
erocs: You could utilize a try-with-resources to guarantee closure of sc upon exiting the block.

challenge_14/java/jdfurlan/src/ReversingALinkedList.java
```diff
@@ -0,0 +1,71 @@
+
+import java.util.Scanner;
+
+public class ReversingALinkedList {
+
+	private static void printList(ListNode head) {
+		while (head != null) {
+			System.out.print(head.data);
```
erocs: This won't print a separator between node values so they will run together. Hopefully all your values end with a space.  ;)

challenge_5/java/jdfurlan/src/FindTheDifference.java
```diff
@@ -0,0 +1,21 @@
+
+public class FindTheDifference {
+
+	public static void main(String[] args) {
+		String s = "abcd";
+		String t = "dbeca";
+
+		boolean[] b = new boolean[128];
```
erocs: This wouldn't work if there were multiple of the same character in the given strings. A simple change to using int to hold counts of the letter occurrences would fix that though.

Also, the problem description states that these will only be lowercase letters (ascii assumed) so you only need an array of 26.

challenge_8/java/jdfurlan/src/RandomPointerLinkedList.java
```diff
@@ -0,0 +1,62 @@
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Random;
+/**
+ * 
+ * @author jesse furlan
+ * 
+ * THIS CODE IS INCOMPLETE
```
erocs: It would be good to have this in the README.md for this challenge. That file gets shown when navigating to the directory in github so the notice would be quite visible to anyone looking.

challenge_9/java/jdfurlan/src/Squares.java
```diff
@@ -0,0 +1,50 @@
...
+		ArrayList<Integer> myList = orderSquares(arr);
+		for (int i : myList) {
+			System.out.print(i + " ");
+		}
+
+	}
+
+	private static ArrayList orderSquares(int[] arr) {
+		ArrayList<Integer> result = new ArrayList<Integer>();
+		ArrayList<Integer> negatives = new ArrayList<Integer>();
```
erocs: Utilize java's LinkedList for the negatives and positives. That way you can use them like a queue. Specifically, adding to the 0 index of the negatives array will cause the array to be recopied every time to shift all the contained elements. A linked list on the other hand would just add a new node to the front with little overhead.

challenge_5/java/jdfurlan/src/FindTheDifference.java
jdfurlan: It does work with multiple of the same char, try apple and lepaip, it will show 'i' look closer at the logic.

challenge_14/java/jdfurlan/src/ReversingALinkedList.java
jdfurlan: Ya the challenge states "The strings in standard input will be space separated." I have to assume expected input.

erocs: Generally separators aren't included in the parsed data. As you say, it's not too big of a deal for this challenge.

challenge_5/java/jdfurlan/src/FindTheDifference.java
erocs: Try s = "a" and t = "aa".

jdfurlan: You're right! I added a fix though it is a bit hacky. In the event that all chars are the same, and given the fact that we know one char HAS to be added to the original string, we can just print any char in the string if we never found a unique char.

PR #447, opened 2017-01-21
C Challenge 2 (Unreviewed)

0x000000AC:
Just realized challenges are asking things to be generic and not hardcoded, will update in the future.

### DIFF
--- a/challenge_2/c/xor-eax-eax/README.md
+++ b/challenge_2/c/xor-eax-eax/README.md
@@ -1,0 +1,40 @@
+# Challenge_2
+###### C compiled with gcc 4.2.1
+
+### 1. Approch to Solving the problem
+
+We assume that we know the list has numbers which only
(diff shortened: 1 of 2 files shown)


challenge_2/c/xor-eax-eax/README.md
```diff
@@ -0,0 +1,40 @@
...
+###### C compiled with gcc 4.2.1
+
+### 1. Approch to Solving the problem
+
+We assume that we know the list has numbers which only
+occur once.  From that assumption, you could go through the
+list and count how many instances of a particular number there are and then return that.
+However, I wanted to do this more generally and did so
+inside inside of a nested for loop.  The outer loop
+goes through the array and the inner loop checks of that number
```
moonpatroller: You could definitely improve the algo to be more efficient than O(n^2).

PR #448, opened 2017-01-21
[Python] Challenge 10 (Pending)

Loran425:


challenge_10/python/loran425/src/matching_brackets.py
```diff
@@ -0,0 +1,37 @@
...
+                elif char is '}':
+                    braces -= 1
+                elif char is ']':
+                    brackets -= 1
+                if braces < 0 or brackets < 0 or parenth < 0:  # Break for out of order designators
+                    return False
+
+    if braces == 0 and brackets == 0 and parenth == 0:
+        return True
+
```
moonpatroller: Will this return False for [{]} ?

Loran425: Ahh I see the concern. It shouldn't as those variables are initialized to 0 on lines 8-10.

PR #450, opened 2017-01-22
[cpp] Challenge 2 (Pending)

dewie102:


challenge_2/cpp/dewie102/README.md
```diff
@@ -0,0 +1,21 @@
+# Reverse a String
```
karanchawla: This challenge is called Single Number.

dewie102: Yes I copied over the file to get the layout and forgot to change it, thank you for pointing that out.

dewie102:
Also for branching I made the mistake of merging my challenge branch into my master instead of onto the yearofprogramming repository, I am trying to fix it as I go.

PR #453, opened 2017-01-23
[Python] Challenge 13 (Unreviewed)

sarcodian:


challenge_13/python/sarcodian/src/challenge_13.py
```diff
@@ -0,0 +1,30 @@
+def int_palin():
+    '''
```
moonpatroller: You could just reverse the number without getting the length and check if that's equal to the original number.

PR #454, opened 2017-01-23
[C++] Challenge 1 (ReadyForMerge)

alexbotello:
Iterates through string input starting from the last element. For every element output it to screen.


ManuelMeraz:
@alexbotello Just add a new line to your output and I can merge it.

PR #458, opened 2017-01-24
[cpp] Challenge 7 (Pending)

rbrtdambrosio:


ManuelMeraz:
- [ ] **Code Correctness**
- [ ] **Code Style**
- [ ] **Code Documentation**
- [ ] **Path Structure**
- [ ] **Merge Pull Request**

challenge_7/cpp/rbrt/src/[cpp]Challenge_7.cpp
```diff
@@ -0,0 +1,50 @@
...
+using namespace std;
+
+
+ int main()
+{
+ 		int n; 
+		cout<< "Please insert the array length N >> ";
+		cin>>n;
+
+		int x[1000]={0};
```
Comment (ManuelMeraz) [inline challenge_7/cpp/rbrt/src/[cpp]Challenge_7.cpp]: What if the user inputs 1001?

challenge_7/cpp/rbrt/src/[cpp]Challenge_7.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+		cout<< "Please insert an array from 0 to N>> ";
```
Comment (ManuelMeraz) [inline challenge_7/cpp/rbrt/src/[cpp]Challenge_7.cpp]: Your job is to generate this list or use the test cases we have. What if the list is of size 1,000,000?

challenge_7/cpp/rbrt/src/[cpp]Challenge_7.cpp
```diff
@@ -0,0 +1,50 @@
...
+			Total= Total-x[i] ;
+	 
+		}
+ 		cout<<Total<<endl;
+ 
+	return 0;
+}
+
+
+int find_max(int x[],int n){
```
Comment (ManuelMeraz) [inline challenge_7/cpp/rbrt/src/[cpp]Challenge_7.cpp]: You shouldn't include functions that you don't use in your solution.

PR #459, opened 2017-01-24
[Go] Challenge 1 (Unreviewed)

rchicoli:


challenge_1/go/rchicoli/reverse.go
```diff
@@ -0,0 +1,10 @@
+package rchicoli
+
+func ReverseString(s string) string {
+	l := len(s) - 1
+	r := make([]byte, l+1)
```
erocs: Go actually uses `rune` for its character type. This works for ASCII text but not for the full range of Unicode characters.

rchicoli: Thx for tip @erocs I've improved the script and added an unicode char for testing this. Now everything passes. Looking forward for the next challenges.

PR #460, opened 2017-01-25
Python challenge 1 (Unreviewed) - Python challenge 0 (Reviewed)

ukirderohit:


karanchawla:
Change title to  Python challenge 0 (Reviewed) and Challenge 1 (Unreviewed)
Read the Project readme to prevent merging multiple challenges in the same PR.

ukirderohit:
Sure Thanks. will take care from challenge 2

erocs:
We also want a README.md to document the solution and how to run it. For Python this is especially useful to differentiate between Python 2 and 3. For example, in these two solutions you use both the Python 2 and 3 form of print.

challenge_1/python/ukirderohit/reverse.py
```diff
@@ -0,0 +1,2 @@
+s = raw_input("Enter String: ")
+print (s[::-1])
```
erocs: Assuming Python2 due to the use of raw_input. This should just be `print s[::-1]`.

PR #467, opened 2017-01-28
[cpp] Challenge 3 (Unreviewed)

dewie102:


dewie102:
I added comments to my code here as well, let me know if you need anything else! I will refer to that as well, thank you!

PR #469, opened 2017-01-29
[cpp] Challenge 4 (Unreviewed)

dewie102:


dewie102:
I commented the code, is that enough to merge it?

PR #474, opened 2017-01-30
[Python] Challenge 4 (Unreviewed)

CindyB:


challenge_4/python/CindyB/src/Challenge4.py
```diff
@@ -0,0 +1,15 @@
+import sys
+
+from Tree import Tree
+from TreeNode import TreeNode
+
+input = [4,2,7,1,3,6,9]
+myTree = Tree(TreeNode(input[0]))
```
erocs: We encourage the use of Python's PEP8 for style and formatting. In this case myTree would instead be my_tree. (https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables)  No need to change this if you don't want to, but going forward it would be nice.

challenge_4/python/CindyB/src/Challenge4.py
```diff
@@ -0,0 +1,15 @@
...
+input.remove(input[0])
+
+for value in input:
+    myTree.insert(value,myTree.GetRoot())
```
erocs: It's odd that inserting to a tree object would require passing in that same tree's root node. This could instead be abstracted away from the caller.

in class Tree:
```
def insert(self, value):
  self._insert(value, self.GetRoot())

def _insert(self, value, node):
  # Same as previous insert, except recursively calling _insert.
```

challenge_4/python/CindyB/src/TreeNode.py
```diff
@@ -0,0 +1,37 @@
+class TreeNode:
+
+    def __init__(self,value):
+        self.value = value
+        self.rightChild = None
+        self.leftChild = None
+
+    def GetValue(self):
```
erocs: Due to how Python and it's properties work, there's no need to create getters / setters to start as they can be created later if necessary without modifying the callers. Instead the class variables should be directly used for such a simple use case.

challenge_4/python/CindyB/src/Tree.py
```diff
@@ -0,0 +1,30 @@
+from TreeNode import TreeNode
+
+class Tree:
+
+    def __init__(self,root):
+        self.root = root
+
+    def invert(self):
```
erocs: Obsolete method which could be removed?

erocs:
- [x] **Code Correctness**
- [ ] **Code Style**
- [x] **Code Documentation**
- [x] **Path Structure**
- [ ] **Merge Pull Request**

CindyB:
Thanks for all the comments ! I'll correct everything and do another commit

erocs:
Typically you'll want to update the title to have (ReadyToMerge), but I'm here so I'll just merge it.

PR #475, opened 2017-01-31
[Python] Challenge 3 (ReadyForReview)

sysek:


erocs:
Update the title to be (ReadyForReview) when you're satisfied with it.  :)

sysek:
updated ;)

PR #485, opened 2017-02-05
[C] Challenge 0 (ReadyForReview)

sysek:


jarydremillard:
Hi sysek,

Can you explain what's happening in majority.py?

